PR #6, opened 2023-06-13
major - v2.0.0

OrNovo:


OrNovo:
we still need to add a releaser and image publishing workflows. we also need to change coralogix.coralogix to coralogix.com before publishing.

PR #11, opened 2023-06-16
[CHORE] adding helm chart

nicolastakashi:
Adding Coralogix Operator Helm Chart.


charts/coralogix-operator/values.yaml
```diff
@@ -0,0 +1,105 @@
...
+
+# -- Coralogix operator container config
+coralogixOperator:
+  # --  Coralogix operator Image
+  image:
+    repository: coralogixrepo/coralogix-operator
+    tag: ""
+
+  # -- Coralogix Account Region
+  region: EUROPE2
```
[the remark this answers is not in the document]
nicolastakashi: Yup!
You're right, forget to remove after test 😄

charts/coralogix-operator/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+name: coralogix-operator
+sources:
+  - https://github.com/coralogix/coralogix-operator
+version: 0.1.0
+appVersion: "0.1.0"
+kubeVersion: ">=1.16.0-0"
+home: https://github.com/coralogix/coralogix-operator
+keywords:
+  - operator
+  - prometheus
```
oded-dd, line 17: Should be Prometheus operator I think

charts/coralogix-operator/templates/secret.yaml
```diff
@@ -0,0 +1,17 @@
...
+  labels:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- with .Values.secret.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+type: Opaque
+stringData:
+  CORALOGIX_API_KEY: {{ .Values.secret.data.apiKey }}
```
oded-dd, line 16: CORALOGIX_REGION is also required no?

charts/coralogix-operator/values.yaml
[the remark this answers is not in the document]
oded-dd: I would add it as ENVVAR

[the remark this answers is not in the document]
nicolastakashi: We're already using as env var, but since it's not a sensible value I didn't require coming from the secret.
But if you think it's better have everything on the secret I can change to.

charts/coralogix-operator/Chart.yaml
nicolastakashi: Makes sense!
I'm going to change

charts/coralogix-operator/templates/secret.yaml
nicolastakashi: since it's not a sensible value I didn't require coming from the secret.
But if you think it's better have everything on the secret I can change to.

oded-dd (approved):

charts/coralogix-operator/templates/secret.yaml
oded-dd: Whatever you think makes more sense

nicolastakashi:
> I noticed that you've duplicated the CRDs from `config/crds/base/`to `charts/coralogix-operator/templates/crds`
> 
> As we make changes to the APIs, are the two folders synced with automation? How do we make sure they stay the same?

Good point!
Let's address that in a dedicated PR, but currently it was done manually.

matej-g (approved):
Looks good 👍

PR #19, opened 2023-06-22
[CHORE] adding helm validation

nicolastakashi:
I'm adding helm install validation for the required fields.


oded-dd (approved):

matej-g:
@povilasv is it possible to validate conditions like this with the JSON schema? IMHO we could have both, seems like e.g. OTEL does the same (they have required `mode` but also error message from `NOTES.txt`.

matej-g (approved):
👍

PR #22, opened 2023-06-22
patch - adding default value for alert timeWindow from PrometheusRule

OrNovo:
PrometheusRule `for` field is optional but in Coralogix it's required.
Another deference, is that in Coralogix it can be just one of - [1m, 5m, 10m, 15m, 20m, 30m, 1H, 2H, 4H, 6H, 12H, 24H].
How do we want to handle empty `for`, and how do we want to handle `for` which is not one of the options above?

### DIFF
--- a/controllers/prometheusrule_controller.go
+++ b/controllers/prometheusrule_controller.go
@@ -265,7 +265,7 @@ func prometheusInnerRuleToCoralogixAlert(prometheusRule prometheus.Rule) coralog
 
 	timeWindow, ok := prometheusAlertForToCoralogixPromqlAlertTimeWindow[prometheusRule.For]
 	if !ok {
-		timeWindow = prometheusAlertForToCoralogixPromqlAlertTimeWindow["10m"]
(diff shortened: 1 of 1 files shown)


controllers/prometheusrule_controller.go
```diff
@@ -263,7 +263,10 @@ func prometheusInnerRuleToCoralogixAlert(prometheusRule prometheus.Rule) coralog
 		notificationPeriod = defaultCoralogixNotificationPeriod
 	}
 
-	timeWindow := prometheusAlertForToCoralogixPromqlAlertTimeWindow[prometheusRule.For]
+	timeWindow, ok := prometheusAlertForToCoralogixPromqlAlertTimeWindow[prometheusRule.For]
```
nicolastakashi, line 266: Ideally users should be able to configure the default options of the operator, through flags

nicolastakashi: Since Prometheus default value is 0 this means as soon as the expression gives us some results it will fire an alert, I would say we may stay as close as possible to this behavior and use the default for as 1m

nicolastakashi (approved):

PR #23, opened 2023-06-23
[CHORE] Removing master name from workflows

nicolastakashi:
I'm removing the `master` branch name from all Github workflows after changing the branch name from `master` to `main`.


matej-g (approved):
🎸

PR #32, opened 2023-07-18
batch - support integration-name instead of integration-ID

OrNovo:
e.g. -

`integrationName: WebhookAlerts` instead of `integrationID: 2235`


controllers/alphacontrollers/alert_controller.go
```diff
@@ -992,8 +995,17 @@ func flattenNotification(notification *alerts.AlertNotification) coralogixv1alph
 
 	switch integration := notification.GetIntegrationType().(type) {
 	case *alerts.AlertNotification_IntegrationId:
-		flattenedNotification.IntegrationID = new(int32)
-		*flattenedNotification.IntegrationID = int32(integration.IntegrationId.GetValue())
+		id := strconv.Itoa(int(integration.IntegrationId.GetValue()))
+		webhookStr, err := coralogixv1alpha1.WebhooksClient.GetWebhook(context.Background(), id)
+		if err != nil {
```
OrNovo, line 1000: not sure how we want to handle APIs errors

controllers/alphacontrollers/alert_controller.go
```diff
@@ -992,8 +995,17 @@ func flattenNotification(notification *alerts.AlertNotification) coralogixv1alph
...
+			return flattenedNotification
+		}
+		var m map[string]interface{}
+		if err = json.Unmarshal([]byte(webhookStr), &m); err != nil {
```
OrNovo, line 1004: Also not sure about this error handling

apis/coralogix/v1alpha1/alert_types.go
```diff
@@ -1053,6 +1061,23 @@ func expandNotification(notification Notification) (*alerts.AlertNotification, e
 	return result, nil
 }
 
+func extractIntegrationID(name string) (uint32, error) {
```
nicolastakashi, line 1064: Not required, but I'd suggest getIntegration, findIntegration or serachIntegration

controllers/alphacontrollers/alert_controller.go
```diff
@@ -992,8 +995,17 @@ func flattenNotification(notification *alerts.AlertNotification) coralogixv1alph
 
 	switch integration := notification.GetIntegrationType().(type) {
 	case *alerts.AlertNotification_IntegrationId:
-		flattenedNotification.IntegrationID = new(int32)
-		*flattenedNotification.IntegrationID = int32(integration.IntegrationId.GetValue())
+		id := strconv.Itoa(int(integration.IntegrationId.GetValue()))
+		webhookStr, err := coralogixv1alpha1.WebhooksClient.GetWebhook(context.Background(), id)
```
nicolastakashi, line 999: Should we avoid context.BackGround()?
And get the context from the caller?

controllers/alphacontrollers/alert_controller.go
```diff
@@ -992,8 +995,17 @@ func flattenNotification(notification *alerts.AlertNotification) coralogixv1alph
 
 	switch integration := notification.GetIntegrationType().(type) {
 	case *alerts.AlertNotification_IntegrationId:
```
nicolastakashi, line 997: Should this still exist, even if we're removing the integration id?

OrNovo: It suppose to be nil when you remove it

OrNovo:
@nicolastakashi I guess charts also have to be updated?

nicolastakashi:
> @nicolastakashi I guess charts also have to be updated?

We need to update the CRDs and also the Image version.
Worth we have a task to automate this soon

OrNovo:
> > @nicolastakashi I guess charts also have to be updated?
> 
> We need to update the CRDs and also the Image version. Worth we have a task to automate this soon

so to leave it to different PR?

apis/coralogix/v1alpha1/alert_types.go
```diff
@@ -1053,6 +1061,23 @@ func expandNotification(notification Notification) (*alerts.AlertNotification, e
 	return result, nil
 }
 
+func searchIntegrationID(ctx context.Context, name string) (uint32, error) {
```
nicolastakashi, line 1064: Just wondering if maybe makes sense for this function to return a integration struct, so in the future, we can leverage this method to manage integrations using the Operator.

OrNovo: When we mange integrations we (probably) won't have to query all the webhooks

controllers/alphacontrollers/alert_controller.go
ArthurSens: If we don't have think kind of knowledge, I guess it is fine to fail the reconciliation

controllers/alphacontrollers/alert_controller.go
ArthurSens: I think flattenNotification should return errors, and the error will be handled in the reconciliation loop

ArthurSens (commented):
Awesome Or, thanks for tackling this one!

I left a few comments regarding error handling. It should be good to merge after that

ArthurSens:
Also food for thought for a future improvement, we could avoid multiple calls to the webhook API if we do a single one when the operator is initializing and cache those webhooks in memory.

The operator itself can provide the webhook IDs, instead of calling the Webhook API once per Alert resource

controllers/alphacontrollers/alert_controller.go
OrNovo: I think it would be better to leave it to reconciliation anyway to avoid complicate logic here

ArthurSens: Yeah, let's just return the error and decide what to do with it during reconciliation

controllers/alphacontrollers/alert_controller.go
OrNovo: The purpose of flatten is to initialize or update the status of the alert. I don't want to fail the update of the status just because of a failure in placing the integration-name. How about leaving the integration-name nil?

ArthurSens: If we leave the integration name as nil, the operator will try to reconcile it again since the Spec and Status are different, right? The loop will continue forever, until we don't fail to fetch a valid `alias`.

I think we want to fail the status update here, it will continue this reconciliation loop, but at least the metric `reconciliation_errors_total` provided by kubebuilder will increase, and we can trigger alerts for this scenario

OrNovo: The reconcile happens forever anyway, the question is just in witch period.
If the Spec and Status are different it will try to update the alert with Update call to the backend but the period to the next reconciliation stays the same.

ArthurSens: Thanks for the explanation :) 

Still, I believe we want to fail here so we can monitor this failures with kubebuilder built-in metrics

OrNovo: It makes the code little bit ugly :'(  but let me know if you think this is a good solution

OrNovo:
> Also food for thought for a future improvement, we could avoid multiple calls to the webhook API if we do a single one when the operator is initializing and cache those webhooks in memory.
> 
> The operator itself can provide the webhook IDs, instead of calling the Webhook API once per Alert resource

TBH I'm not a big fan of handling states inside the operator, but I guess we won't have a choice.

controllers/alphacontrollers/alert_controller.go
```diff
@@ -212,11 +216,17 @@ func (r *AlertReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 		log.V(1).Info("Alert was updated", "alert", jstr)
 	}
 
+	if flattenErr != nil {
+		log.Error(err, "Received an error while flattened Alert")
```
ArthurSens, line 220:
```suggestion
		log.Error(flattenErr, "Received an error while flattened Alert")
```

controllers/alphacontrollers/alert_controller.go
```diff
@@ -954,35 +964,50 @@ func flattenDaysOfWeek(daysOfWeek []alerts.DayOfWeek, daysOffset int32) []coralo
...
+
+	if len(errors) > 0 {
+		var errStr string
+		for _, err := range errors {
+			errStr += err.Error() + "\n"
+		}
+		return result, fmt.Errorf(errStr)
+	}
+
+	return result, nil
```
ArthurSens, line 984: nit:

Just in case you think this would be better, there are a few libraries that deal with concatenating multiple errors, for example:
* https://pkg.go.dev/github.com/hashicorp/go-multierror
* https://pkg.go.dev/errors

We could get rid of this if/for in case it sounds useful to you

ArthurSens (approved):
Awesome Or! Thanks for tackling this so quickly

I've added just a few more comments, but they are non-blocking. Feel free to merge or to address them, it is up to you :)

controllers/alphacontrollers/alert_controller.go
ArthurSens: We could also move this upwards, and return the error even before we try to call Status.Update

OrNovo: I prefer the status to be updated with nil integration-name

PR #38, opened 2023-07-30
patch - saving external IDs immediately

OrNovo:


OrNovo:
> No tests updated / added ? How do we know if it works?

@povilasv The only way that I can think of to test it, is to cause a failure on the `flatten` step (expand the CR `status` after the resource was create in Coralogix), and checking if the resources are created over and over again.
So I tested it manually and it looks fine, but do you have better idea?

PR #40, opened 2023-08-02
Upgrade controller-runtime and enable pprof

ArthurSens:
Now that we're deploying the operator in clusters with bigger scales, we want to understand where our resources are being spent.

This PR upgrades controller-runtime to 1.15 so we can use the new pprof feature (https://github.com/kubernetes-sigs/controller-runtime/pull/1943)


matej-g (approved):
👌

PR #43, opened 2023-08-03
Fixing notification groups deep equal bug

OrNovo:
It turns out that NotificationGroups and Notifications within each NotificationGroup do not maintain the creation order, therefore a different type of comparison between spec and status is needed.

### DIFF
--- a/apis/coralogix/v1alpha1/alert_types.go
+++ b/apis/coralogix/v1alpha1/alert_types.go
@@ -1197,20 +1197,59 @@ func DeepEqualNotificationGroups(notificationGroups []NotificationGroup, actualN
 		}
 	}
 
-	for i := range notificationGroups {
-		notificationGroup, actualNotificationGroup := notificationGroups[i], actualNotificationGroups[i]
-		{
-			if equal, diff := notificationGroup.DeepEqual(actualNotificationGroup); !equal {
-				return false, utils.Diff{
-					Name:    fmt.Sprintf("Notifications.%d.%s", i, diff.Name),
-					Desired: diff.Desired,
-					Actual:  diff.Actual,
+	desiredGroupByFieldsToNotification := getGroupByFieldsToNotificationMap(notificationGroups)
+	actualGroupByFieldsToNotification := getGroupByFieldsToNotificationMap(actualNotificationGroups)
+	for groupByFields, notifications := range desiredGroupByFieldsToNotification {
+		if actualNotifications, ok := actualGroupByFieldsToNotification[groupByFields]; !ok {
+			return false, utils.Diff{
+				Name:    fmt.Sprintf("Notifications.gropup-by:%s", groupByFields),
+				Desired: notifications,
+				Actual:  nil,
+			}
+		} else {
+			desiredNotificationsByIntegrationName := getNotificationsByIntegrationNameMap(notifications)
+			actualNotificationsByIntegrationName := getNotificationsByIntegrationNameMap(actualNotifications)
+
+			for integrationName, notification := range desiredNotificationsByIntegrationName {
+				if actualNotification, ok := actualNotificationsByIntegrationName[integrationName]; !ok {
+					return false, utils.Diff{
+						Name:    fmt.Sprintf("Notifications.gropup-by:%s.%s", groupByFields, integrationName),
+						Desired: notification,
+						Actual:  nil,
+					}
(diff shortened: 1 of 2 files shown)


apis/coralogix/v1alpha1/alert_types.go
```diff
@@ -1197,20 +1197,59 @@ func DeepEqualNotificationGroups(notificationGroups []NotificationGroup, actualN
...
+	desiredGroupByFieldsToNotification := getGroupByFieldsToNotificationMap(notificationGroups)
+	actualGroupByFieldsToNotification := getGroupByFieldsToNotificationMap(actualNotificationGroups)
+	for groupByFields, notifications := range desiredGroupByFieldsToNotification {
+		if actualNotifications, ok := actualGroupByFieldsToNotification[groupByFields]; !ok {
+			return false, utils.Diff{
+				Name:    fmt.Sprintf("Notifications.gropup-by:%s", groupByFields),
+				Desired: notifications,
+				Actual:  nil,
+			}
+		} else {
```
ArthurSens, line 1209: This else statement could be removed to reduce one indentation level, making the code easier to read :)

https://dev.to/zachsnoek/clean-up-your-code-by-removing-unneeded-indentation-3g9i

nicolastakashi (approved):

ArthurSens (commented):
> therefore a different type of comparison between spec and status is needed.

Could you elaborate on what is this "different type of comparison"? It's a bit hard to understand just by reading the code

And nice find discovering that the API doesn't return notification groups ordered. I'm not sure if your solution works as expected though, [maps are not ordered in golang](https://www.educative.io/answers/how-to-iterate-over-a-golang-map-in-sorted-order), so I'm not sure if using a map is helping here



One easy way to make sure that your function is working is by adding a unit test :)

OrNovo:
> > therefore a different type of comparison between spec and status is needed.
> 
> Could you elaborate on what is this "different type of comparison"? It's a bit hard to understand just by reading the code
> 
> And nice find discovering that the API doesn't return notification groups ordered. I'm not sure if your solution works as expected though, [maps are not ordered in golang](https://www.educative.io/answers/how-to-iterate-over-a-golang-map-in-sorted-order), so I'm not sure if using a map is helping here
> 
> One easy way to make sure that your function is working is by adding a unit test :)

Notifications-Groups are unique by their group-by arrays.
Also Notifications are unique by their integration-name/id

So I create map of group-by->[]Notification and compare, then, for the Notification list I create map of integration->Notification and compare.

apis/coralogix/v1alpha1/alert_types.go
OrNovo: But then I'll need to declare on actualNotifications and ok outside the if-else (line 1203).
What do you prefer?

ArthurSens: The if remains there, just the else that goes away:

```go
		if actualNotifications, ok := actualGroupByFieldsToNotification[groupByFields]; !ok {
			return false, utils.Diff{
				Name:    fmt.Sprintf("Notifications.gropup-by:%s", groupByFields),
				Desired: notifications,
				Actual:  nil,
			}
		} 

		desiredNotificationsByIntegrationName := getNotificationsByIntegrationNameMap(notifications)
		actualNotificationsByIntegrationName := getNotificationsByIntegrationNameMap(actualNotifications)

		for integrationName, notification := range desiredNotificationsByIntegrationName {
			if actualNotification, ok := actualNotificationsByIntegrationName[integrationName]; !ok {
				return false, utils.Diff{
					Name:    fmt.Sprintf("Notifications.gropup-by:%s.%s", groupByFields, integrationName),
					Desired: notification,
					Actual:  nil,
				}
			} else if equal, diff := notification.DeepEqual(actualNotification); !equal {
				return false, utils.Diff{
					Name:    fmt.Sprintf("Notifications.gropup-by:%s.%s.%s", groupByFields, integrationName, diff.Name),
					Desired: diff.Desired,
					Actual:  diff.Actual,
				}
			}
		}
```

ArthurSens (approved):
So sorry for being annoying, and thanks for your patience and quick resolution ❤️

OrNovo:
> So sorry for being annoying, and thanks for your patience and quick resolution ❤️

NP, I'll talk with Oded about prioritize adding tests. Currently focusing on adding validations.

PR #69, opened 2023-08-28
fix: coralogix operator rulegroup tests

povilasv:
Adds:
- unit tests
- improved error handling
- error wrapping
- get rids of panics for rulegroups

References: ES-68

### DIFF
--- a/controllers/alphacontrollers/rulegroup_controller.go
+++ b/controllers/alphacontrollers/rulegroup_controller.go
@@ -78,6 +78,7 @@ func (r *RuleGroupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{RequeueAfter: defaultErrRequeuePeriod}, err
 	}
 
+	//TODO: rename finalizer
 	// name of our custom finalizer
 	myFinalizerName := "batch.tutorial.kubebuilder.io/finalizer"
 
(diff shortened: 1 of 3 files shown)


controllers/alphacontrollers/rulegroup_controller.go
```diff
@@ -124,25 +125,43 @@ func (r *RuleGroupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		if err != nil {
+			log.Error(err, "Error marshalling rule group", "Name", ruleGroupCRD.Name, "Namespace", ruleGroupCRD.Namespace)
+			return ctrl.Result{RequeueAfter: defaultErrRequeuePeriod}, err
+		}
+
 		log.V(1).Info("Creating Rule-Group", "ruleGroup", jstr)
 		if createRuleGroupResp, err := rulesGroupsClient.CreateRuleGroup(ctx, createRuleGroupReq); err == nil {
-			jstr, _ := jsm.MarshalToString(createRuleGroupResp)
+			jstr, err := jsm.MarshalToString(createRuleGroupResp)
+			if err != nil {
```
OrNovo, line 161: I'm not sure if we want to fail in a case of marshalling error for two reasons -
1. The marshalling is just for logs purpose and doesn't affect the actual logic of the operator.
2. In this specific case, returning an error (even though the creation succeed) will cause infinit recreation of resources in coralogix.

controllers/alphacontrollers/rulegroup_controller.go
```diff
@@ -175,14 +199,18 @@ func (r *RuleGroupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 			log.Error(err, "Received an error while updating a Rule-Group", "ruleGroup", updateRuleGroupReq)
 			return ctrl.Result{RequeueAfter: defaultErrRequeuePeriod}, err
 		}
-		jstr, _ := jsm.MarshalToString(updateRuleGroupResp)
+		jstr, err := jsm.MarshalToString(updateRuleGroupResp)
+		if err != nil {
+			log.Error(err, "Error marshalling rule group", "Name", ruleGroupCRD.Name, "Namespace", ruleGroupCRD.Namespace)
+			return ctrl.Result{RequeueAfter: defaultErrRequeuePeriod}, err
```
OrNovo, line 205: Same issue here

OrNovo (approved):

PR #79, opened 2023-09-05
fix: helm charts improve installation validation

povilasv:
FIX ES-79

region and secret data are mandatory, so adding more validation during helm install

Tested manually using helm:

```
helm upgrade --install test . --set secret.data.apiKey=...
```

```
Release "test" does not exist. Installing it now.

Error: execution error at (coralogix-operator/templates/NOTES.txt:10:3): [ERROR] 'coralogixOperator.region' must be set. Please select one of APAC1,APAC2,EUROPE1,EUROPE2,USA1,USA2 regions. See https://coralogix.com/docs/coralogix-domain/ for more information.
```


matej-g (approved):
Nice 👍

PR #113, opened 2024-07-09
[REFACTOR] prometheus rules reconcile

nicolastakashi:
Adding WithEventFilter to allow us reduce the amount of resources we're reconciling, if the object didn't match the reconcile rule it will not be added to the work queue.

REf: https://stuartleeks.com/posts/kubebuilder-event-filters-part-1-delete/


OrNovo (approved):
Great!

PR #121, opened 2024-07-31
outbound-webhooks

OrNovo:


celaus:
IAC-22

nicolastakashi:
I'm going to review it between today and tomorrow 🙏🏽

controllers/alphacontrollers/outboundwebhook_controller.go
```diff
@@ -0,0 +1,453 @@
...
+	client.Client
+	OutboundWebhooksClient clientset.OutboundWebhooksClientInterface
+	Scheme                 *runtime.Scheme
+}
+
+//+kubebuilder:rbac:groups=coralogix.coralogix.com,resources=outboundwebhooks,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=coralogix.coralogix.com,resources=outboundwebhooks/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=coralogix.coralogix.com,resources=outboundwebhooks/finalizers,verbs=update
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
```
nicolastakashi, line 53: we can remove this comment block

controllers/alphacontrollers/outboundwebhook_controller.go
```diff
@@ -0,0 +1,453 @@
...
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.4/pkg/reconcile
+var (
+	outboundWebhookFinalizerName = "outbound-webhook.coralogix.com/finalizer"
+)
+
+func (r *OutboundWebhookReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	var (
+		resultError = ctrl.Result{RequeueAfter: 40}
+		resultOk    = ctrl.Result{RequeueAfter: defaultErrRequeuePeriod}
```
nicolastakashi, line 69: Let's not requeu if reconcile has sucess, we removed if from others controllers to leverage the sync period of kube builde

controllers/alphacontrollers/outboundwebhook_controller.go
```diff
@@ -0,0 +1,453 @@
...
+}
+
+func (r *OutboundWebhookReconciler) create(ctx context.Context, log logr.Logger, webhook *coralogixv1alpha1.OutboundWebhook) error {
+	createRequest, err := webhook.ExtractCreateOutboundWebhookRequest()
+	if err != nil {
+		log.Error(err, fmt.Sprintf("Error to extract create-request out of the outbound-webhook -\n%v", webhook))
+		return err
+	}
+
+	log.V(1).Info(fmt.Sprintf("Creating outbound-webhook-\n%s", protojson.Format(createRequest)))
```
nicolastakashi, line 130:
```suggestion
	log.V(int(zapcore.DebugLevel)).Info(fmt.Sprintf("Creating outbound-webhook-\n%s", protojson.Format(createRequest)))
```

controllers/alphacontrollers/outboundwebhook_controller.go
```diff
@@ -0,0 +1,453 @@
...
+	createResponse, err := r.OutboundWebhooksClient.CreateOutboundWebhook(ctx, createRequest)
+	if err != nil {
+		log.Error(err, fmt.Sprintf("Received an error while creating outbound-webhook -\n%s", protojson.Format(createRequest)))
```
nicolastakashi, line 133: Let's only log on the higher function, here we should always return an error message.

```suggestion
return fmt.Errorf("received an error while trying to createoutbound-webhook CRD: %w", err)
```

Since we're adding the following code on the reconcile function

```
log := log.FromContext(ctx).WithValues(
		"outboundWebhook", req.NamespacedName.Name,
		"namespace", req.NamespacedName.Namespace,
	)
```

All logs will contain the webhook name and name space

controllers/alphacontrollers/outboundwebhook_controller.go
```diff
@@ -0,0 +1,453 @@
...
+		return err
+	}
+	log.Info(fmt.Sprintf("outbound-webhook was created-\n%s", protojson.Format(createResponse)))
+
+	webhook.Status = coralogixv1alpha1.OutboundWebhookStatus{
+		ID:                  ptr.To(createResponse.Id.GetValue()),
```
nicolastakashi, line 139: I don't think we should store this fields on status, only id.
This fields are already present on the spec and we can get if from there if needed

controllers/alphacontrollers/outboundwebhook_controller.go
OrNovo: This is more informative, because it shows the call for the backend. This way we can easily track where is the issue.

controllers/alphacontrollers/outboundwebhook_controller.go
OrNovo: I don't have solid opinion about it yet, but in this case some of the fields are "computed" (will be determined by the backend), so it can be helpful to keep also the status.

oded-dd (approved):

PR #129, opened 2024-08-07
fixing alert-controller bugs and add tests

OrNovo:


OrNovo:
@nicolastakashi I decided not to delete the status fields at the moment. The change are only related to alerts, the rest of the files are small typo changes.

apis/coralogix/v1alpha1/alert_types.go
```diff
@@ -1004,9 +1009,23 @@ func expandNotificationGroups(ctx context.Context, notificationGroups []Notifica
 	return result, nil
 }
 
-func expandNotificationGroup(ctx context.Context, notificationGroup NotificationGroup) (*alerts.AlertNotificationGroups, error) {
+func getWebhooksNamesToIds(ctx context.Context, log logr.Logger) (map[string]uint32, error) {
+	webhooksNamesToIds := make(map[string]uint32)
+	log.V(1).Info("Listing all outgoing webhooks")
+	listWebhooksResp, err := WebhooksClient.ListAllOutgoingWebhooks(ctx, &webhooks.ListAllOutgoingWebhooksRequest{})
+	if err != nil {
+		log.Error(err, "Failed to list all outgoing webhooks")
```
nicolastakashi, line 1017: we're logging too much, I don't think we need to log it here, let's just return a formated error and only log on the controller.
```suggestion
		return nil, fmt.Errorf("Failed to list all outgoing webhooks, err, err)
```

oded-dd (approved):

PR #151, opened 2024-09-22
Add official cluster shorthands to region flags

assafad1:
Also ran `make generate-api-docs` and `make helm-update-crds`.


charts/coralogix-operator/templates/crds/coralogix.com_recordingrulegroupsets.yaml
celaus, line 1: Should those be commited? They are going to change again next time

celaus (approved):

PR #157, opened 2024-09-26
Support custom domain

assafad1:


main.go
```diff
@@ -104,12 +107,32 @@ func main() {
...
+		if !slices.Contains(validRegions, region) {
+			err := fmt.Errorf("region value is '%s', but can be one of %q", region, validRegions)
+			setupLog.Error(err, "invalid arguments for running operator")
+			os.Exit(1)
+		}
+		targetUrl = regionToGrpcUrl[region]
+	}
+
+	if domain != "" {
+		targetUrl = fmt.Sprintf("ng-api-grpc.%s:443", domain)
```
lootag, line 133: You're reassigning the variable here, therefore ignroing the previous `if` block. It works, but I think it would be a bit clearer if you used an `if ... else` construct, because it would evident that only one of these conditions affects the value `target url`

lootag (approved):

lootag (approved):

PR #210, opened 2024-11-13
Clean up after failing to store the status

assafad1:
The issue: 
When the user creates a CR (CRD instance), the operator creates a remote resource and stores its ID in the CR status. If this status update operation fails, the ID property remains empty, and then the operator will create the same remote resource in the next reconciliation, so there is a duplicated remote resource.
The solution:
If the status update operation fails, delete the newly created remote resource. The next reconciliation will take care of recreating it.

### DIFF
--- a/internal/controller/coralogix/outboundwebhook_controller.go
+++ b/internal/controller/coralogix/outboundwebhook_controller.go
@@ -128,6 +128,9 @@ func (r *OutboundWebhookReconciler) create(ctx context.Context, log logr.Logger,
 		ID: ptr.To(createResponse.Id.GetValue()),
 	}
 	if err = r.Status().Update(ctx, webhook); err != nil {
+		if err := r.deleteRemoteWebhook(ctx, log, webhook.Status.ID, r.OutboundWebhooksClient); err != nil {
+			return fmt.Errorf("error to delete outbound-webhook after status update error -\n%v", webhook)
+		}
 		return fmt.Errorf("error to update outbound-webhook status -\n%v", webhook)
 	}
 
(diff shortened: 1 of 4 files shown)


internal/controller/coralogix/outboundwebhook_controller.go
```diff
@@ -229,3 +229,13 @@ func (r *OutboundWebhookReconciler) delete(ctx context.Context, log logr.Logger,
 
 	return nil
 }
+
+func (r *OutboundWebhookReconciler) deleteRemoteWebhook(ctx context.Context, log logr.Logger, webhookID *string, client clientset.OutboundWebhooksClientInterface) error {
+	log.V(int(zapcore.DebugLevel)).Info("Deleting outbound-webhook from remote", "id", webhookID)
+	if _, err := client.Delete(ctx, &cxsdk.DeleteOutgoingWebhookRequest{Id: wrapperspb.String(*webhookID)}); err != nil && status.Code(err) != codes.NotFound {
+		return fmt.Errorf("error to delete outbound-webhook -\n%v", webhookID)
```
celaus, line 236: I think printing the error message here is more useful (id + message)

internal/controller/coralogix/rulegroup_controller.go
```diff
@@ -156,6 +148,10 @@ func (r *RuleGroupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 			id := createRuleGroupResp.GetRuleGroup().GetId().GetValue()
 			ruleGroupCRD.Status = coralogixv1alpha1.RuleGroupStatus{ID: &id}
 			if err := r.Status().Update(ctx, ruleGroupCRD); err != nil {
+				if err := r.deleteRemoteRuleGroup(ctx, log, ruleGroupCRD.Status.ID); err != nil {
+					log.Error(err, "Error on deleting RecordingRuleGroupSet after status update error", "Name", ruleGroupCRD.Name, "Namespace", ruleGroupCRD.Namespace)
```
celaus, line 152: `log.V(1).Error` ?

celaus (approved):

internal/controller/coralogix/rulegroup_controller.go
assafad1: Seems like this big Reconcile func uses `log.Error` for error. We need to handle this logging differences in another PR.

PR #221, opened 2024-11-26
Fix RuleGroup unit test

assafad1:
Fix RuleGroup unit test which #186 broke.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ vet: ## Run go vet against code.
 
 .PHONY: unit-tests
 unit-tests: manifests generate fmt vet envtest ## Run tests.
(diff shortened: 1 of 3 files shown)


config/webhook/manifests.yaml
```diff
@@ -9,7 +9,7 @@ webhooks:
   clientConfig:
     service:
       name: webhook-service
-      namespace: coralogix-operator-system
+      namespace: system
```
OrNovo, line 12: Isn't `system` a dedicated namespace for k8s stuff?

assafad1: kustomize adds a prefix to all namespaces fields, see this for example:
https://github.com/coralogix/coralogix-operator/blob/a970bcce9a2f979ea066f5846139fd4a00643410/config/manager/manager.yaml#L18 it doesn't have the full `coralogix-operator-system` as well. But I removed this change from this PR, maybe will include it in the cert-manager one.

OrNovo (approved):

PR #236, opened 2024-12-03
Fix PrometheusRule issue

### LINKED ISSUE #213: PrometheusRule invalid alert configuration shouldn't block other alerts
When applying a PrometheusRule with multiple alerts, if one of the alerts is misconfigured, the whole PrometheusRule reconciliation is stopped and the remaining alerts are not created, even if they are valid. The operator should block the creation of the invalid alerts only, and allow other alerts to be created.

assafad1:
Fixes #213

### DIFF
--- a/internal/controller/prometheusrule_controller.go
+++ b/internal/controller/prometheusrule_controller.go
@@ -16,6 +16,7 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
 	"strings"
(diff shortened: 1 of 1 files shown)


internal/controller/prometheusrule_controller.go
```diff
@@ -174,6 +174,7 @@ func (r *PrometheusRuleReconciler) convertPrometheusRuleAlertToCxAlert(ctx conte
 	}
 
 	alertsToKeep := make(map[string]bool)
+	var errorsEncountered []error
```
OrNovo, line 177: Consider using [errors](https://pkg.go.dev/errors) instead

OrNovo (approved):

PR #264, opened 2025-02-06
fix: check if alert exists in remote

GoshaCor:
If alert was deleted manually with clickops, the CRD status alertID will fail to create a alert because it will try to update existing one.
This PR extends the controller to check if alerts exists..

### DIFF
--- a/internal/controller/coralogix/v1beta1/alert_controller.go
+++ b/internal/controller/coralogix/v1beta1/alert_controller.go
@@ -74,7 +74,12 @@ func (r *AlertReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 		return ctrl.Result{RequeueAfter: utils.DefaultErrRequeuePeriod}, err
 	}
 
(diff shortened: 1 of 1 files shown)


assafad1:
Hi @GoshaCor, I'm not sure this change is needed. You are right that once a remote alert is deleted, it will try to update existing one. But once the update fails with `NotFound` error, we are removing ID from status. This triggers another reconciliation in which a new alert will be created.
If you encountered any issues with alerts not being recreated, we might have a bug somewhere, but these proposed changes shouldn't be the solution.

GoshaCor:
> Hi @GoshaCor, I'm not sure this change is needed. You are right that once a remote alert is deleted, it will try to update existing one. But once the update fails with `NotFound` error, we are removing ID from status. This triggers another reconciliation in which a new alert will be created. If you encountered any issues with alerts not being recreated, we might have a bug somewhere, but these proposed changes shouldn't be the solution.

PR #266, opened 2025-02-10
Refactor to use a common reconciler

assafad1:


cmd/main.go
```diff
@@ -21,6 +21,7 @@ import (
 	"os"
 	"strings"
 
+	coralogixcontroller "github.com/coralogix/coralogix-operator/internal/controller/coralogix"
```
assafad1, line 24: - The imports paths become really unclear with this. Can we create a `coralogix-reconciler` package?

go.mod
```diff
@@ -46,6 +45,7 @@ require (
 	github.com/go-task/slim-sprig/v3 v3.0.0 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
+	github.com/golang/protobuf v1.5.4 // indirect
```
assafad1, line 48: please run `go mod tidy`. I will add validation for it.

internal/controller/coralogix/coralogix_reconciler.go
```diff
@@ -0,0 +1,143 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/coralogix/coralogix-operator/internal/utils"
+)
+
+var Client client.Client
+var Schema *runtime.Scheme
```
assafad1, line 36: I think we should avoid using global vars, it can add some pitfalls like race conditions, difficult debugging, and unexpected state changes, especially in concurrent program like the operator.

internal/controller/coralogix/coralogix_reconciler.go
```diff
@@ -0,0 +1,143 @@
...
+	return nil
+}
+
+func objToGVK(obj client.Object) string {
+	gvks, _, _ := Schema.ObjectKinds(obj)
+	if len(gvks) == 0 {
+		return ""
+	}
+	return gvks[0].String()
+}
```
assafad1, line 143: why not using `obj.GetObjectKind().GroupVersionKind()` like in other places?

internal/controller/coralogix/coralogix_reconciler.go
```diff
@@ -0,0 +1,143 @@
...
+		controllerutil.AddFinalizer(obj, r.FinalizerName())
+		if err := Client.Update(ctx, obj); err != nil {
+			return fmt.Errorf("error updating %s: %w", obj.GetObjectKind().GroupVersionKind(), err)
+		}
+	}
+	return nil
+}
+
+func RemoveFinalizer(ctx context.Context, log logr.Logger, obj client.Object, r CoralogixReconciler) error {
+	log.V(1).Info(fmt.Sprintf("Removing finalizer from %s", obj.GetObjectKind().GroupVersionKind()))
```
assafad1, line 129: I think `fmt.Sprintf` is redundant, if so please fix for all additions.

internal/controller/coralogix/coralogix_reconciler.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+		return ctrl.Result{RequeueAfter: utils.DefaultErrRequeuePeriod}, err
+	}
+
+	if !r.CheckIDInStatus(obj) {
+		log.V(1).Info("Resource ID is missing; handling creation for resource")
+		if createdObj, err := r.HandleCreation(ctx, log, obj); err != nil {
+			log.Error(err, "Error handling creation")
+			return ctrl.Result{RequeueAfter: utils.DefaultErrRequeuePeriod}, err
+		} else if err = Client.Status().Update(ctx, createdObj); err != nil {
```
assafad1, line 65: redundant `else`. If the first condition is met, you return.

internal/controller/coralogix/coralogix_reconciler.go
```diff
@@ -0,0 +1,143 @@
...
+			return ctrl.Result{RequeueAfter: utils.DefaultErrRequeuePeriod}, err
+		}
+
+		return ctrl.Result{}, nil
+	}
+
+	if !utils.GetLabelFilter().Matches(obj.GetLabels()) {
+		log.V(1).Info("Resource labels do not match label filter; handling deletion")
+		if err := r.HandleDeletion(ctx, log, obj); err != nil {
+			log.Error(err, "Error deleting from remote")
```
assafad1, line 96: change the log to `"Error handling deletion"`.

internal/controller/coralogix/coralogix_reconciler.go
assafad1: More importantly, I think you forgot the part where we delete remote resource if status update failed.

tests/e2e/global_router_test.go
```diff
@@ -27,6 +28,7 @@ var _ = Describe("GlobalRouter", Ordered, func() {
 
 	BeforeAll(func() {
 		crClient = ClientsInstance.GetControllerRuntimeClient()
+		coralogix.Client = crClient
```
assafad1, line 31: same

tests/e2e/group_test.go
```diff
@@ -48,6 +49,7 @@ var _ = Describe("Group", Ordered, func() {
 
 	BeforeEach(func() {
 		crClient = ClientsInstance.GetControllerRuntimeClient()
+		coralogix.Client = crClient
```
assafad1, line 52: same

tests/e2e/recording_rule_group_test.go
```diff
@@ -42,6 +42,7 @@ var _ = Describe("RecordingRuleGroupSet", Ordered, func() {
 
 	BeforeAll(func() {
 		crClient = ClientsInstance.GetControllerRuntimeClient()
+		coralogix.Client = crClient
```
assafad1, line 45: same

internal/controller/coralogix/coralogix_reconciler.go
OrNovo: it doesn't work without using the schema

internal/controller/coralogix/coralogix_reconciler.go
OrNovo: I don't agree, it's like initialising the manager  in the main for instance

internal/controller/coralogix/coralogix_reconciler.go
OrNovo: This way createdObj lives only in the if/else scope

internal/controller/coralogix/coralogix_reconciler.go
assafad1: it's not like the manager initialization. There we do:
```go
mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), mgrOpts)
```
and then we pass it to controllers with:
```go
if err = (&v1alpha1controllers.GroupReconciler{
CXClientSet: clientSet,
Client:      mgr.GetClient(),
Scheme:      mgr.GetScheme(),
}).SetupWithManager(mgr); err != nil {
setupLog.Error(err, "unable to create controller", "controller", "Group")
os.Exit(1)
}
```

or to webhooks with:
```go
if err = webhookcoralogixv1alpha1.SetupOutboundWebhookWebhookWithManager(mgr); err != nil {
	setupLog.Error(err, "unable to create webhook", "webhook", "OutboundWebhook")
	os.Exit(1)
}
```

internal/controller/coralogix/coralogix_reconciler.go
assafad1: But why do you need it in this scope only?
You continue to operate on the resource also out of the scope of this if/else.. better to override the obj so you can always operate on the latest resource version.

internal/controller/coralogix/coralogix_reconciler.go
assafad1: but other places have `obj.GetObjectKind().GroupVersionKind().Kind` - this is why I asked. Let's make it work if possible, no reason to reimplement existing utils.

OrNovo: Where do we have obj.GetObjectKind().GroupVersionKind().Kind?

internal/controller/coralogix/coralogix_reconciler.go
assafad1: No. I'm not sure why can't we continue to have them stored in the reconcilers like today. Anyway we are passing reconcilers to `ReconcileResource()`. And some reconcilers need them in their code, e.g. group needs it for referencing scope and custom rule, etc.
let's treat `ReconcileResource()` as a util and not store critical info as global vars in its package.

internal/controller/coralogix/coralogix_reconciler.go
assafad1: You have `obj.GetObjectKind().GroupVersionKind()` 4 times in your code.

internal/controller/coralogix/coralogix_reconciler.go
OrNovo: I think the less code a new reconciler needs to add, the better and client and schema are unnecessary for most of the reconcilers.

assafad1: You are only saving 2 lines per controller file, for some of the controllers. but if you want to do it let's use in the `/api` folder. We use the client there. Also let's change it so there won't be any global vars, but package scopes vars that we init with a func:
```go
var client client.Client

func InitClient(mgrClient) {
  client = mgrClient
}
```
and same for scheme. This is less error prone, because it's clearer when you override it.

api/coralogix/v1alpha1/globalrouter_types.go
```diff
@@ -25,6 +25,7 @@ import (
 
 	cxsdk "github.com/coralogix/coralogix-management-sdk/go"
 
+	coralogix_reconciler "github.com/coralogix/coralogix-operator/internal/controller/coralogix/coralogix-reconciler"
```
assafad1, line 28:
```suggestion
	coralogixreconciler "github.com/coralogix/coralogix-operator/internal/controller/coralogix/coralogix-reconciler"
```

this follows the same pattern as other imports, e.g:
```
coralogixv1alpha1 "github.com/coralogix/coralogix-operator/api/coralogix/v1alpha1"
```

cmd/main.go
```diff
@@ -21,6 +21,7 @@ import (
 	"os"
 	"strings"
 
+	coralogixcontroller "github.com/coralogix/coralogix-operator/internal/controller/coralogix/coralogix-reconciler"
```
assafad1, line 24: import renaming and location are wrong

cmd/main.go
```diff
@@ -259,88 +256,69 @@ func main() {
...
 		NotificationsClient: clientSet.Notifications(),
-		Client:              mgr.GetClient(),
-		Scheme:              mgr.GetScheme(),
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "Preset")
 		os.Exit(1)
 	}
 	if err = (&v1alpha1controllers.GlobalRouterReconciler{
 		NotificationsClient: clientSet.Notifications(),
 		Client:              mgr.GetClient(),
```
assafad1, line 321: why do you leave the k8s client?

cmd/main.go
```diff
@@ -259,88 +256,69 @@ func main() {
...
 		OutboundWebhooksClient: clientSet.Webhooks(),
-		Client:                 mgr.GetClient(),
-		Scheme:                 mgr.GetScheme(),
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "OutboundWebhook")
 		os.Exit(1)
 	}
 	if err = (&v1alpha1controllers.ApiKeyReconciler{
 		ApiKeysClient: clientSet.APIKeys(),
 		Client:        mgr.GetClient(),
```
assafad1, line 265: if you decided to make the k8s client global, please be consistent and treat it this way. No reason to move it around.

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
```diff
@@ -0,0 +1,162 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/coralogix/coralogix-operator/internal/utils"
+)
+
+var (
+	k8sClient client.Client
+	schema    *runtime.Scheme
```
assafad1, line 37:
```suggestion
	scheme    *runtime.Scheme
```

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
```diff
@@ -0,0 +1,162 @@
...
+
+func InitClient(c client.Client) {
+	k8sClient = c
+}
+
+func GetClient() client.Client {
+	return k8sClient
+}
+
+func InitSchema(s *runtime.Scheme) {
```
assafad1, line 48:
```suggestion
func InitScheme(s *runtime.Scheme) {
```

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
```diff
@@ -0,0 +1,162 @@
...
+	HandleDeletion(ctx context.Context, log logr.Logger, obj client.Object) error
+	FinalizerName() string
+	CheckIDInStatus(obj client.Object) bool
+}
+
+func ReconcileResource(ctx context.Context, req ctrl.Request, obj client.Object, r CoralogixReconciler) (ctrl.Result, error) {
+	gvk := objToGVK(obj)
+	log := log.FromContext(ctx).WithValues(
+		"gvk", gvk,
+		"name", req.NamespacedName.Name, "namespace", req.NamespacedName.Namespace)
```
assafad1, line 65:
```suggestion
	log := log.FromContext(ctx).WithValues(
		"gvk", gvk,
		"name", req.NamespacedName.Name, 
		"namespace", req.NamespacedName.Namespace)
```

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
```diff
@@ -0,0 +1,162 @@
...
+	if err := k8sClient.Get(ctx, req.NamespacedName, obj); err != nil {
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{RequeueAfter: utils.DefaultErrRequeuePeriod}, err
+	}
+
+	if !r.CheckIDInStatus(obj) {
+		log.V(1).Info("Resource ID is missing; handling creation for resource")
+		var err error
```
assafad1, line 76: please initialize the error at the top of the function and use it in the get as well

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
```diff
@@ -0,0 +1,162 @@
...
+			log.Error(err, "Error deleting from remote")
+			return ctrl.Result{RequeueAfter: utils.DefaultErrRequeuePeriod}, err
+		}
+		return ctrl.Result{}, nil
+	}
+
+	log.V(1).Info("Handling update")
+	if err := r.HandleUpdate(ctx, log, obj); err != nil {
+		if cxsdk.Code(err) == codes.NotFound {
+			log.V(1).Info(fmt.Sprintf("%s not found on remote, recreating it", gvk))
```
assafad1, line 122: `fmt.Sprintf` is redundant I think. Also here we are not recreating it, but only remove id from status. I think it's misleading.

internal/controller/coralogix/v1alpha1/outboundwebhook_controller_test.go
```diff
@@ -194,22 +196,22 @@ func TestOutboundWebhookUpdate(t *testing.T) {
...
+				//		Config: &cxsdk.GenericWebhook{
+				//			GenericWebhook: &cxsdk.GenericWebhookConfig{
+				//				Uuid:    wrapperspb.String("updated-uuid"),
+				//				Method:  cxsdk.GenericWebhookConfigPost,
+				//				Headers: map[string]string{"updated-key": "updated-value"},
+				//				Payload: wrapperspb.String("updated-payload"),
+				//			},
+				//		},
+				//	},
+				//}, nil)
```
assafad1, line 214: uncomment?

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
```diff
@@ -0,0 +1,162 @@
...
+
+func RemoveFinalizer(ctx context.Context, log logr.Logger, obj client.Object, r CoralogixReconciler) error {
+	gvk := objToGVK(obj)
+	log.V(1).Info("Removing finalizer", "gvk", gvk)
+	controllerutil.RemoveFinalizer(obj, r.FinalizerName())
+	if err := k8sClient.Update(ctx, obj); err != nil {
+		return fmt.Errorf("error updating %s: %w", gvk, err)
+	}
+	return nil
+}
```
assafad1, line 162: please locate right under AddFinalizer func

assafad1:
I think you have a bug in the `HandleUpdate()` logic. In the `ReconcileResource()` you do:
```
if err := r.HandleUpdate(ctx, log, obj); err != nil {
  if cxsdk.Code(err) == codes.NotFound
```

but in `HandleUpdate()` you do:
```
updateResponse, err := r.NotificationsClient.ReplaceConnector(ctx, updateRequest)
 if err != nil {
   return fmt.Errorf("error on updating remote connector: %w", err)
 }
``` 

I think it won't interpret this error code as `NotFound`, but as `Unknown`, then we won't remove id from status, and the resource won't be recreated.

If needed please fix this, and make sure resources are being recreated if deleted from API/UI.

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
OrNovo: fmt.Sprintf is for adding the gvk in the beginning of the message

internal/controller/coralogix/v1alpha1/outboundwebhook_controller_test.go
OrNovo: I'm no longer doing a reading request in updating webhooks, this is unnecessary

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
assafad1: but why do you need it? you built the log with this info already. Other logs doesn't have it.

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
```diff
@@ -0,0 +1,168 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package coralogixreconciler
+
+import (
+	"context"
+	"fmt"
+
+	cxsdk "github.com/coralogix/coralogix-management-sdk/go"
```
assafad1, line 21: import location

OrNovo (approved):

PR #269, opened 2025-02-12
Remove severities from TCOTracesPolicies

assafad1:


celaus (dismissed):

charts/coralogix-operator/templates/crds/coralogix.com_globalrouters.yaml
```diff
@@ -140,6 +141,7 @@ spec:
                             type: object
                         required:
                         - connector
+                        - preset
```
nicolastakashi, line 144: Should this be part of the PR?

celaus (approved):

PR #276, opened 2025-02-20
Improve error handling

OrNovo:


api/coralogix/utils.go
```diff
@@ -152,3 +153,23 @@ func WrapperspbStringToStringPointer(s *wrapperspb.StringValue) *string {
...
+	RemoteUpdatedSuccessfully = "RemoteUpdatedSuccessfully"
+	RemoteUpdateFailed        = "RemoteUpdateFailed"
+	RemoteDeletionFailed      = "RemoteDeletionFailed"
+	RemoteDeletedSuccessfully = "RemoteDeletedSuccessfully"
+	RemoteResourceNotFound    = "RemoteResourceNotFound"
+	InternalK8sError          = "InternalK8sError"
+
+	ConditionTypeError        = "Error"
+	ConditionTypeRemoteSynced = "RemoteSynced"
+)
```
assafad1, line 169: Let's put all of this in a new file under `internal/utils/conditions.go`. It's going to be used in other places like metrics so it's relevant for all the repo.

api/coralogix/utils.go
```diff
@@ -152,3 +153,23 @@ func WrapperspbStringToStringPointer(s *wrapperspb.StringValue) *string {
...
+
+// ConditionsAware represents a CRD type that has been enabled with metav1.Conditions, it can then benefit of a series of utility methods.
+type ConditionsAware interface {
```
assafad1, line 172: Nice. 
I would rename it to `ConditionsObj` as it's going to be implemented by objects. 
Also, it's not a util. I would also locate it in a new file `api/conditions_obj.go` or something.

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
```diff
@@ -71,49 +77,66 @@ func ReconcileResource(ctx context.Context, req ctrl.Request, obj client.Object,
...
 		log.V(1).Info("Resource ID is missing; handling creation for resource")
+		if conditionsAware, ok := (obj).(api.ConditionsAware); ok {
+			conditions := conditionsAware.GetConditions()
+			syncingCondition := metav1.Condition{
+				Type:    api.ConditionTypeRemoteSynced,
+				Status:  metav1.ConditionFalse,
+				Reason:  api.RemoteResourceNotFound,
+				Message: "Syncing remote resource",
+			}
+			if meta.SetStatusCondition(&conditions, syncingCondition) {
```
assafad1, line 93: I think we should wrap all of this with functions that will be located in zinternal/utils/conditions.goz. They would look like:
```go
func SetSyncedConditionFalse(conditions, reason, message){
 ...
}

func SetSyncedConditionTrue(conditions, reason, message){
  ...
}

...
```

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
```diff
@@ -160,6 +183,68 @@ func RemoveFinalizer(ctx context.Context, log logr.Logger, obj client.Object, r
...
+	if conditionsAware, updateStatus := (obj).(api.ConditionsAware); updateStatus {
+		conditions := conditionsAware.GetConditions()
+		meta.SetStatusCondition(&conditions, metav1.Condition{
+			Type:    api.ConditionTypeRemoteSynced,
+			Status:  metav1.ConditionTrue,
+			Reason:  reason,
+			Message: "Remote resource synced",
+		})
+		meta.RemoveStatusCondition(&conditions, api.ConditionTypeError)
+		conditionsAware.SetConditions(conditions)
```
assafad1, line 238: this should be in `internal/utils/conditions.go` as well, which maybe makes this func redundant?

api/coralogix/utils.go
```diff
@@ -152,3 +153,23 @@ func WrapperspbStringToStringPointer(s *wrapperspb.StringValue) *string {
 	}
 	return &s.Value
 }
+
+const (
+	RemoteCreatedSuccessfully = "RemoteCreatedSuccessfully"
+	RemoteCreateFailed        = "RemoteCreateFailed"
```
assafad1, line 159:
```suggestion
	RemoteCreationFailed        = "RemoteCreationFailed"
```

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
OrNovo: I find it a bit too specific (implementing setting a specific condition to true or false)

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
OrNovo: I'm not sure, cause the the way we implement the conditions is by the end depends on the controller and not a general behavior.

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
assafad1: having func per true/false is not a must.. just my personal preference. But you can do:
```go
func SetSyncedCondition(status bool, conditions, reason, message string) conditions {
            syncingCondition := metav1.Condition{
				Type:    conditionTypeRemoteSynced,
				Status:  status,
				Reason:  reason,
				Message: message,
	    }
	    if meta.SetStatusCondition(&conditions, syncingCondition)}
                         ...
}
```

The important part for is to build and set the condition in the `conditions.go` file util. To make the controller as short and readable as possible

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
```diff
@@ -160,6 +177,49 @@ func RemoveFinalizer(ctx context.Context, log logr.Logger, obj client.Object, r
 	return nil
 }
 
+func ManageErrorWithRequeue(ctx context.Context, log logr.Logger, obj client.Object, reason string, err error) (reconcile.Result, error) {
+	requeueAfter := utils.DefaultErrRequeuePeriod
```
assafad1, line 181: it's never used, since it's always overriden in the if/else. I think you should start with: 
`requeueAfter := utils. MinErrRequeuePeriod` and remove the else.

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
```diff
@@ -160,6 +177,49 @@ func RemoveFinalizer(ctx context.Context, log logr.Logger, obj client.Object, r
...
+			conditionsObj.SetConditions(conditions)
+			if err := GetClient().Status().Update(context, obj); err != nil {
+				log.Error(err, "unable to update status")
+				return reconcile.Result{RequeueAfter: utils.DefaultErrRequeuePeriod}, err
+			}
+		}
+	}
+
+	return reconcile.Result{}, nil
+}
```
assafad1, line 221: I think this func is problematic. It does a few things that don't share the same input, and its purpose isn't clear from its name. What about having a dedicated func for calculating requeue duration, something like (could be written nicer, don't judge it too much):
```go
func calculateErrorRequeuePeriod(obj client.Object) time.Duration {
	requeueAfter := utils.DefaultErrRequeuePeriod
	if conditionsObj, ok := (obj).(utils.ConditionsObj); ok {
		requeueAfter = utils.MinErrRequeuePeriod
		if errCondition := meta.FindStatusCondition(conditionsObj.GetConditions(), utils.ConditionTypeError); errCondition != nil && errCondition.Status == metav1.ConditionTrue {
			lastRequeuePeriod := metav1.Now().Sub(errCondition.LastTransitionTime.Time).Round(time.Second)
			return time.Duration(
				math.Max(float64(utils.MinErrRequeuePeriod),
					math.Min(float64(utils.MaxErrRequeuePeriod), float64(lastRequeuePeriod.Nanoseconds()*2)),
				),
			)
		}
	}

	return requeueAfter
}
```

and then another func that will be located in `conditions.go` that updates the error condition (same as you did for synced condition).

OrNovo:
```go
func ManageErrorWithRequeue(ctx context.Context, log logr.Logger, obj client.Object, reason string, err error) (reconcile.Result, error) {
	requeueAfter := utils.DefaultErrRequeuePeriod
	if conditionsObj, ok := (obj).(utils.ConditionsObj); ok {
		// if the error condition is already set, we increase the requeue period
		conditions := conditionsObj.GetConditions()
		requeueAfter = calculateErrorRequeuePeriod(conditions)

		if utils.SetSyncedConditionFalse(&conditions, obj.GetGeneration(), reason, err.Error()) {
			conditionsObj.SetConditions(conditions)
			if err = k8sClient.Status().Update(ctx, obj); err != nil {
				log.Error(err, "unable to update status")
				return reconcile.Result{RequeueAfter: requeueAfter}, err
			}
		}
	}

	return reconcile.Result{RequeueAfter: requeueAfter}, err
}

func calculateErrorRequeuePeriod(conditions []metav1.Condition) time.Duration {
	if errCondition := meta.FindStatusCondition(conditions, utils.ConditionTypeError); errCondition != nil && errCondition.Status == metav1.ConditionTrue {
		lastRequeuePeriod := metav1.Now().Sub(errCondition.LastTransitionTime.Time).Round(time.Second)
		return time.Duration(
			math.Max(float64(utils.MinErrRequeuePeriod),
				math.Min(float64(utils.MaxErrRequeuePeriod), float64(lastRequeuePeriod.Nanoseconds()*2)),
			),
		)
	}
	return utils.MinErrRequeuePeriod
}
```

I'll change it to this code, as I want to extract conditions from obj only once.

config/manager/kustomization.yaml
```diff
@@ -10,4 +10,4 @@ patchesStrategicMerge:
 images:
 - name: controller
   newName: coralogix-operator-image
-  newTag: latest
+  newTag: v0.0.1
```
assafad1, line 13: remove

internal/utils/conditions.go
```diff
@@ -0,0 +1,54 @@
+package utils
```
assafad1, line 1: missing license

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
```diff
@@ -164,6 +181,52 @@ func RemoveFinalizer(ctx context.Context, log logr.Logger, obj client.Object, r
 	return nil
 }
 
+func ManageErrorWithRequeue(ctx context.Context, log logr.Logger, obj client.Object, reason string, err error) (reconcile.Result, error) {
```
assafad1, line 184: missing setting the error condition?

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
```diff
@@ -75,39 +80,50 @@ func ReconcileResource(ctx context.Context, req ctrl.Request, obj client.Object,
...
-		return ctrl.Result{RequeueAfter: utils.DefaultErrRequeuePeriod}, err
+		return ManageErrorWithRequeue(ctx, log, obj, utils.ReasonInternalK8sError, err)
 	}
 
 	if !r.CheckIDInStatus(obj) {
 		log.V(1).Info("Resource ID is missing; handling creation for resource")
-		if obj, err = r.HandleCreation(ctx, log, obj); err != nil {
+		if ConditionsObj, ok := (obj).(utils.ConditionsObj); ok {
+			conditions := ConditionsObj.GetConditions()
+			if utils.SetSyncedConditionFalse(&conditions, obj.GetGeneration(), utils.ReasonRemoteResourceNotFound, "Syncing remote resource") {
```
assafad1, line 90: shouldn't we do it at the top of the func? if the func was called, it means the resource isn't synced, no?

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
```diff
@@ -164,6 +181,52 @@ func RemoveFinalizer(ctx context.Context, log logr.Logger, obj client.Object, r
...
 
+func ManageErrorWithRequeue(ctx context.Context, log logr.Logger, obj client.Object, reason string, err error) (reconcile.Result, error) {
+	requeueAfter := utils.DefaultErrRequeuePeriod
+	if conditionsObj, ok := (obj).(utils.ConditionsObj); ok {
+		// if the error condition is already set, we increase the requeue period
+		conditions := conditionsObj.GetConditions()
+		requeueAfter = calculateErrorRequeuePeriod(conditions)
+
+		utils.SetErrorConditionTrue(&conditions, (obj).GetGeneration(), reason, err.Error())
+		utils.SetSyncedConditionFalse(&conditions, (obj).GetGeneration(), reason, err.Error())
```
assafad1, line 192: I think setting false for the synced condition here is redundant. We need to do it at the top of the reconcile, and here to only set the error condition to true.

tests/e2e/alert_test.go
```diff
@@ -151,11 +154,21 @@ var _ = Describe("Alert", Ordered, func() {
...
+
+			if len(fetchedAlert.Status.Conditions) != 1 {
+				return fmt.Errorf("alert status conditions length is not 1, got %d", len(fetchedAlert.Status.Conditions))
+			}
+
+			if !meta.IsStatusConditionTrue(fetchedAlert.Status.Conditions, utils.ConditionTypeRemoteSynced) {
+				return fmt.Errorf("alert status condition %s is not true", utils.ConditionTypeRemoteSynced)
+			}
+
+			return nil
```
assafad1, line 170: You put it under `By("Fetching the Alert ID")`, can you move it to be after?

tests/e2e/alert_test.go
```diff
@@ -191,6 +204,136 @@ var _ = Describe("Alert", Ordered, func() {
 		}, time.Minute, time.Second).Should(Equal(codes.NotFound))
 	})
 
+	It("Should store err condition in status", func(ctx context.Context) {
+		By("Creating Alert")
```
assafad1, line 208:
```suggestion
		By("Creating Alert with a non-existing webhook")
```

tests/e2e/alert_test.go
```diff
@@ -191,6 +204,136 @@ var _ = Describe("Alert", Ordered, func() {
...
+		fetchedAlert := &coralogixv1beta1.Alert{}
+		Eventually(func(g Gomega) error {
+			g.Expect(crClient.Get(ctx, types.NamespacedName{Name: alert.Name, Namespace: alert.Namespace}, fetchedAlert)).To(Succeed())
+
+			if len(fetchedAlert.Status.Conditions) != 2 {
+				return fmt.Errorf("alert status conditions length is not 2, got %d", len(fetchedAlert.Status.Conditions))
+			}
+
+			if !meta.IsStatusConditionTrue(fetchedAlert.Status.Conditions, utils.ConditionTypeError) {
+				return fmt.Errorf("alert status condition %s is not true", utils.ConditionTypeError)
```
assafad1, line 283:
```suggestion
			Expect(meta.IsStatusConditionTrue(fetchedAlert.Status.Conditions, utils.ConditionTypeError)).To(BeTrue())

```

tests/e2e/alert_test.go
```diff
@@ -191,6 +204,136 @@ var _ = Describe("Alert", Ordered, func() {
...
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Fetching the Alert")
+		fetchedAlert := &coralogixv1beta1.Alert{}
+		Eventually(func(g Gomega) error {
+			g.Expect(crClient.Get(ctx, types.NamespacedName{Name: alert.Name, Namespace: alert.Namespace}, fetchedAlert)).To(Succeed())
+
+			if len(fetchedAlert.Status.Conditions) != 2 {
+				return fmt.Errorf("alert status conditions length is not 2, got %d", len(fetchedAlert.Status.Conditions))
+			}
```
assafad1, line 280:
```suggestion
			Expect(fetchedAlert.Status.Conditions).To(HaveLen(2))
```

tests/e2e/alert_test.go
assafad1: Please do for all

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
assafad1: Also, the RemoteSynced condition shouldn't include the error.. For that we have the Error condition.

tests/e2e/alert_test.go
```diff
@@ -147,15 +149,17 @@ var _ = Describe("Alert", Ordered, func() {
...
-		Eventually(func(g Gomega) error {
+		Eventually(func(g Gomega) {
 			g.Expect(crClient.Get(ctx, types.NamespacedName{Name: alertName, Namespace: testNamespace}, fetchedAlert)).To(Succeed())
-			if fetchedAlert.Status.ID != nil {
-				alertID = *fetchedAlert.Status.ID
-				return nil
-			}
-			return fmt.Errorf("Alert ID is not set")
+
+			g.Expect(meta.IsStatusConditionTrue(fetchedAlert.Status.Conditions, utils.ConditionTypeRemoteSynced))
```
assafad1, line 157:
```suggestion
			g.Expect(meta.IsStatusConditionTrue(fetchedAlert.Status.Conditions, utils.ConditionTypeRemoteSynced)).To(BeTrue())
```

tests/e2e/alert_test.go
```diff
@@ -191,6 +195,117 @@ var _ = Describe("Alert", Ordered, func() {
...
+
+		err := crClient.Create(ctx, newAlert)
+		Expect(err).ToNot(HaveOccurred())
+
+		By("Fetching the Alert and verifying the RemoteSynced condition is false")
+		fetchedAlert := &coralogixv1beta1.Alert{}
+		Eventually(func(g Gomega) {
+			g.Expect(crClient.Get(ctx, types.NamespacedName{Name: newAlert.Name, Namespace: newAlert.Namespace}, fetchedAlert)).To(Succeed())
+
+			g.Expect(meta.IsStatusConditionFalse(fetchedAlert.Status.Conditions, utils.ConditionTypeRemoteSynced))
```
assafad1, line 269:
```suggestion
			g.Expect(meta.IsStatusConditionFalse(fetchedAlert.Status.Conditions, utils.ConditionTypeRemoteSynced)).To(BeTrue())
```

assafad1 (dismissed):

api/coralogix/v1alpha1/scope_types.go
```diff
@@ -92,6 +92,16 @@ func (s *ScopeSpec) ExtractScopeFilters() ([]*cxsdk.ScopeFilter, error) {
 // ScopeStatus defines the observed state of Scope.
 type ScopeStatus struct {
 	ID *string `json:"id"`
```
assafad1, line 94: should add `omitempty`?

PR #292, opened 2025-03-10
adding conditions to api-key

OrNovo:


internal/controller/coralogix/v1alpha1/apikey_controller.go
```diff
@@ -18,6 +18,8 @@ import (
 	"context"
 	"fmt"
 
+	coralogixreconciler "github.com/coralogix/coralogix-operator/internal/controller/coralogix/coralogix-reconciler"
+	"github.com/coralogix/coralogix-operator/internal/utils"
```
assafad1, line 22: import location

PR #294, opened 2025-03-10
Add AlertScheduler CRD

assafad1:
- Including CRD docs, validation webhhok and tests.
- Also change Preset and Connector names in e2e tests to be unique (backend requires it now).


api/coralogix/v1alpha1/alertscheduler_types.go
```diff
@@ -0,0 +1,504 @@
...
+	// Alert Scheduler meta labels. Conflicts with `alerts`.
+	// +optional
+	MetaLabels []MetaLabel `json:"metaLabels,omitempty"`
+
+	// Alert references. Conflicts with `metaLabels`.
+	// +optional
+	Alerts []AlertRef `json:"alerts,omitempty"`
+}
+
+type AlertRef struct {
```
OrNovo, line 72: Did you mean to expand it to support backendID?

assafad1: Not the moment. If we'll get a specific request for that, we can add it, because I designed so we can extend it without breaking the CRD.
IMO we shouldn't support references to explicit IDs in backend, as it's not the "k8s way". With explicit backend ID, once this ID changes, all the AlertSchedulers become invalid.

assafad1: BTW, it's the same for Group CRD -> CustomRoles and Scopes.

api/coralogix/v1alpha1/alertscheduler_types.go
```diff
@@ -0,0 +1,505 @@
...
+type AlertSchedulerSpec struct {
+	// Alert Scheduler name.
+	Name string `json:"name"`
+
+	// Alert Scheduler description.
+	// +optional
+	Description string `json:"description,omitempty"`
+
+	// Alert Scheduler enabled. If set to `false`, the alert scheduler will be disabled. True by default.
+	// +optional
```
OrNovo, line 43: How do you set it to true by default?

api/coralogix/v1alpha1/alertscheduler_types.go
```diff
@@ -0,0 +1,505 @@
...
+	Enabled bool `json:"enabled,omitempty"`
+
+	// Alert Scheduler meta labels.
+	// +optional
+	MetaLabels []MetaLabel `json:"metaLabels,omitempty"`
```
OrNovo, line 48: you have metaLabels at `spec` and at `spec.Filter`. Is that on purpose?

api/coralogix/v1alpha1/alertscheduler_types.go
OrNovo: But this way it's required to manage the alerts with k8s. I think we should support both ways.

api/coralogix/v1alpha1/alertscheduler_types.go
```diff
@@ -0,0 +1,505 @@
...
+
+	// The timezone of the time frame. For example, `UTC-4` or `UTC+10`.
+	// +kubebuilder:validation:Pattern=`^UTC[+-]\d{1,2}$`
+	Timezone string `json:"timezone"`
+}
+
+type Recurring struct {
+	// Recurring always.
+	// +optional
+	Always *Always `json:"always,omitempty"`
```
OrNovo, line 123: unless there's a plan to expand it, I would change it to a boolean or even omit this field, but this is only a personal taste.

api/coralogix/v1alpha1/alertscheduler_types.go
assafad1: True. I expect anyone who manages AlertScheduler from k8s, to manage the alerts this way as well. Using an explicit ID doesn't sound very useful to me, as IDs change and then all resources that refer it are broken. 
Anyway, as I mentioned, in case you are right and we'll get such request, the design allows to extend it. We can add it so I don't think it should be a blocker for introducing the CRD.

api/coralogix/v1alpha1/alertscheduler_types.go
assafad1: yes, both are needed.

api/coralogix/v1alpha1/alertscheduler_types.go
assafad1: I think we should follow the API here (as I did). If they extend it someday, it will break the CRD.
Also using it with:
```
always: {}
```
is as easy as 
```
always: true
```
and leaving empty may surprise users, so I don't think it should be considered.

api/coralogix/v1alpha1/alertscheduler_types.go
assafad1: added, thanks

OrNovo (approved):

PR #297, opened 2025-03-11
remove omitempty from arrays and adding validation for length where required

OrNovo:
remove omitempty from arrays and adding validation for length where required


api/coralogix/v1alpha1/alert_types.go
```diff
@@ -606,21 +606,21 @@ type TracingFilters struct {
...
 
 	// +optional
-	Services []string `json:"services,omitempty"`
+	Services []string `json:"services"`
 }
 
 type TagFilter struct {
 	Field  string   `json:"field,omitempty"`
-	Values []string `json:"values,omitempty"`
+	Values []string `json:"values"`
```
assafad1, line 623: @OrNovo I think this change makes this field required, which can break existing users, wdyt?

api/coralogix/v1beta1/alert_types.go
```diff
@@ -476,7 +476,7 @@ type ResourceRef struct {
 }
 
 type ActiveOn struct {
-	DayOfWeek []DayOfWeek `json:"dayOfWeek,omitempty"`
+	DayOfWeek []DayOfWeek `json:"dayOfWeek"`
```
assafad1, line 479: same

celaus (approved):

PR #311, opened 2025-03-19
Add a debugging CLI tool

assafad1:
This PR adds a CLI tool named  `cxo-observer` that collects Kubernetes resources related to the Coralogix Operator installation, including both core operator components and custom resources managed by the operator.

### DIFF
--- a/tools/cxo-observer/README.md
+++ b/tools/cxo-observer/README.md
@@ -1,0 +1,77 @@
+# cxo-observer
+
+## Overview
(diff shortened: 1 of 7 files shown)


tools/cxo-observer/README.md
```diff
@@ -0,0 +1,77 @@
+# cxo-observer
```
celaus, line 1: Can you add that for any and all support requests the affected resources from the output has to be attached

celaus (dismissed):

celaus (approved):

PR #315, opened 2025-03-23
Remove explicit CRDs lists

assafad1:
Instead of maintaining CRDs string list for metrics collection and reconcile intervals, use the kinds that are registered in scheme.


charts/coralogix-operator/values.yaml
```diff
@@ -130,9 +130,6 @@ coralogixOperator:
     tcoLogsPolicies: ""
     tcoTracesPolicies: ""
     integration: ""
-    connector: ""
-    preset: ""
-    globalRouter: ""
```
assafad1, line 135: Removing these because these CRDs are behind FF for now.

celaus (approved):

PR #358, opened 2025-05-13
Add labels for chart templates

dleinov:


assafad1 (changes requested):
Hi @dleinov, thanks for your PR. Please see inline comments.

charts/coralogix-operator/templates/deployment.yaml
```diff
@@ -18,6 +18,9 @@ spec:
         {{- end }}
       labels:
         {{- include "coralogixOperator.selectorLabels" . | nindent 8 }}
+        {{- if .Values.podLabels }}
+        {{- toYaml .Values.podLabels | nindent 8 }}
+        {{- end }}
```
assafad1, line 23: Can we follow the same pattern used in:
https://github.com/coralogix/coralogix-operator/blob/main/charts/coralogix-operator/templates/secret.yaml#L6-L9

charts/coralogix-operator/templates/serviceaccount.yaml
```diff
@@ -5,6 +5,9 @@ metadata:
   name: {{ include "coralogixOperator.serviceAccountName" . }}
   labels:
     {{- include "coralogixOperator.labels" . | nindent 4 }}
+    {{- range $key, $value := .Values.serviceMonitor.additionalLabels }}
+    {{ $key }}: {{ $value | quote }}
+    {{- end }}
```
assafad1, line 10: Same comment about pattern. And you applied service monitor labels in the service account file.

charts/coralogix-operator/values.yaml
```diff
@@ -6,6 +6,12 @@ fullnameOverride: ""
 
 imagePullSecrets: []
 
+# -- Custom labels to add into metadata
+additionalLabels: {}
+
+# -- Pod labels for Coralogix operator
+podLabels: {}
+
```
assafad1, line 14: Since it's used in the deployment resource, I think we should include it in the deployment values.

charts/coralogix-operator/values.yaml
```diff
@@ -31,6 +37,9 @@ serviceMonitor:
   # -- Specifies whether a service monitor should be created.
   create: true
 
+  # -- Additional labels to add for ServiceMonitor 
+  additionalLabels: {}
```
assafad1, line 41: Secret is using [`labels`](https://github.com/coralogix/coralogix-operator/blob/872175accb011510008d6301fcee49c2598800e0/charts/coralogix-operator/values.yaml#L63), not `additionalLabels`. Can we align them?

assafad1 (commented):
Hi @dleinov, thanks for your PR. Please see inline comments.

assafad1 (approved):

assafad1:
@dleinov Please sign commits and CLA.

PR #372, opened 2025-05-26
Reduce API server list requests

assafad1:
It turns out that by default, `unstructured.Unstructured` objects are not read from the client cache. 
In the [metrics collector](https://github.com/coralogix/coralogix-operator/blob/main/internal/monitoring/collectors.go#L119-L124) we list all Coralogix custom resources, which caused a huge amount of direct API server List requests.

This PR enables caching for unstructured objects as well, to reduce the amount of direct calls.


doryer (approved):

OrNovo:
This PR improves Prometheus metric collection by:

* Adding **per-GVK caching** with invalidation on resource changes or TTL expiry.
* Using **`sync.Map`** for thread-safe cache access.
* Introducing an **instrumented Kubernetes client** that:

  * Auto-invalidates affected GVKs on Create/Update/Delete.
  * Tracks kube-api usage via a new metric:
    `cx_operator_kube_api_requests_total{verb, resource}`

These changes reduce redundant API calls, improve metric freshness, and increase observability.

PR #423, opened 2025-08-10
Iac 980 - Adding a status column when getting results

NoyaItzhaki:
This PR is implementing the addition of a Status column when getting a resource via kubectl.
Brfore:
`➜  coralogix-operator git:(main) ✗ k get <some-cx-custom-resource>`                                             
`NAMESPACE   NAME           AGE`
`some-ns     some-sample   2s` 

After:
`➜  coralogix-operator git:(main) ✗ k get <some-cx-custom-resource> `                                            
`NAMESPACE   NAME          AGE   STATUS`
`some-ns     some-sample   2s    Synced or Unsynced`


NoyaItzhaki:
@assafad1 
I hope I did the e2e test addition correctly, either way, would love to have a session with you about these e2e tests.
as for the update for the conditioning, I have changed it to be triggered only when necessary, but left the two separate (conditions & status), to prevent extra updates that may occur if only one of the conditions is met.

PR #424, opened 2025-09-07
Fix router test

assafad1:


tests/e2e/global_router_test.go
```diff
@@ -65,31 +65,9 @@ var _ = Describe("GlobalRouter", Ordered, func() {
 		}, time.Minute, time.Second).Should(Succeed())
 	})
 
-	It("Should be updated successfully", func(ctx context.Context) {
```
NoyaItzhaki, line 68: @celaus Don't we want to have that test on a different field instead of just deleting the update method?

NoyaItzhaki (approved):

PR #426, opened 2025-09-15
Add docs sync

celaus:


.github/workflows/dispatch-event-on-docs-change.yml
```diff
@@ -0,0 +1,24 @@
+name: "[Docs] Event for Sync"
+
+on:
+  workflow_dispatch:
+  push:
```
assafad1, line 5: I think we should sync docs on tags (releases).
At the moment docs include unreleased stuff. For example, SLO is in https://coralogix.com/docs/external/coralogix-operator/documents/api/, although isn't available to customers.
@celaus @veronikaambar wdyt?

assafad1 (approved):

PR #495, opened 2026-03-25
feat(outboundwebhook): add IBM Event Notifications webhook support

smitthakkar96:
## Context

`OutboundWebhook` reconciliation already assumes IBM Event Notifications support, but users had no way to configure that webhook type in the CRD (for example, the IBM branch in `outboundwebhook_controller.go`). This PR adds CRD/API support for configuring IBM Event Notifications webhooks so the full flow is actually usable.

This also aligns the operator with Coralogix API behavior, where IBM Event Notifications is documented as a supported outgoing webhook type: [Create an outgoing webhook](https://docs.coralogix.com/api-reference/latest/outgoing-webhooks-service/create-an-outgoing-webhook).


povilasv:
@codex review

povilasv (commented):
Could you add tests for this?

smitthakkar96:
@povilasv I noticed following about our current test setup:

- `kuttl-test.yaml` only lists `tests/integration/prometheusrules`
- `make integration-tests` runs `kubectl kuttl test`, so it follows that suite config
- I also don’t see any CI workflow explicitly targeting `tests/integration/outboundwebhooks`

We do have [`tests/e2e/outbound_webhook_test.go`](https://github.com/coralogix/coralogix-operator/blob/main/tests/e2e/outbound_webhook_test.go), which only creates outbound webhooks for PagerDuty.

Given that, what direction do you prefer for testing these changes:

- Add more E2E coverage per integration type (similar to PagerDuty), or
- Keep E2E minimal and add controller-level tests with a mocked Coralogix API client, so we just validate reconciliation logic

My inclination is the second option for long-term maintainability, but happy to align with your preference.

povilasv:
Second option sounds fine

PR #508, opened 2026-05-05
feat(integration): support sourcing parameters from Kubernetes Secrets

nicolastakashi:
## Why

Several managed `Integration` types carry sensitive material in `spec.parameters` — `gcp-metrics-collector` ships a service account JSON key, `aws-metrics-collector` ships an IAM role ARN that some teams treat as sensitive, `microsoft-365-collector` and `salesforce-events-collector` carry tokens, and so on.

Today, because `IntegrationSpec.Parameters` is a `runtime.RawExtension` that's forwarded verbatim, the only way to set these is to embed the cleartext value directly in the manifest. From there it lands in Git, in rendered Helm output, in `etcd`, in CI logs, and in any `kubectl get -o yaml` an operator runs.

The `Connector` CRD already solves this with `secretKeyRef` on each `ConnectorConfigField`. This PR closes the equivalent gap on `Integration`.

## What

Adds a new optional field `parametersFromSecret` on `IntegrationSpec`:

```yaml
spec:
  integrationKey: gcp-metrics-collector
  version: 1.0.0
  parameters:
    IntegrationName: gcp-metrics-collector-prod
    MetricPrefixes: [cloudsql.googleapis.com]
    ApplicationName: gcp
  parametersFromSecret:
    ServiceAccountKey:
      name: gcp-metrics-collector-sa-key
      key: key.json
```

It's a `map[string]corev1.SecretKeySelector`. At reconcile time the controller resolves each entry against a Kubernetes Secret in the Integration's namespace and merges the resulting string into the parameter set forwarded to the Coralogix API.

`spec.parameters` is now itself optional — manifests where every parameter is sensitive can omit the inline block entirely and source everything from `parametersFromSecret`.

## Design choice

Considered three shapes:

1. **Sibling map** (this PR): keep `parameters` as the schemaless `RawExtension`, add `parametersFromSecret` alongside.
2. **Per-field struct list** matching Connector exactly (`{ fieldName, value | secretKeyRef }`): consistent across CRDs but breaks every existing Integration manifest and loses the schemaless flexibility that lets the operator support new `integrationKey` values without code changes.
3. **`parametersFrom` envFrom-style bulk import**: heavy for the actual problem shape (typically 1 of N parameters is sensitive).

Picked (1) because it's strictly additive, preserves the schemaless design that makes Integration easy to extend, and stays ergonomic for the common all-inline case. The Connector pattern's main benefit — explicit per-field validation — is preserved here via the runtime collision check below.

## Behavior

- **Back-compat**: existing CRs that don't use `parametersFromSecret` are unchanged. The field is optional.
- **Composition**: `parameters` and `parametersFromSecret` are merged. The same key in both is rejected with a clear error during reconciliation.
- **Type scope**: only string-valued parameters can come from a Secret (Secret bytes decode to string). Numeric, boolean, and list-valued params must remain inline. This covers all sensitive fields I'm aware of in current integrations.
- **Optional refs**: `corev1.SecretKeySelector.Optional` is honored. When `optional: true`, a missing Secret or missing key is silently skipped (matching kubelet `envFrom` semantics) and the Integration is created/updated without that parameter. Other read errors (RBAC, transient API failures) still propagate so reconciliation retries. Skips are logged at V(1).
- **RBAC footprint**: same as `Connector` — `get` on `Secrets` in the Integration's namespace.
- **Refactor**: `ExtractCreateIntegrationRequest` / `ExtractUpdateIntegrationRequest` moved from `*IntegrationSpec` to `*Integration` receivers, taking `ctx`. Mirrors `Connector.ExtractConnector(ctx)`. The only call sites are in this repo's controller, both updated.

## Testing

- Unit tests in `api/coralogix/v1alpha1` cover the merge, key-collision, required-error, and optional-skip paths (including both missing-secret and missing-key variants).
- Added an e2e test `Integration with ParametersFromSecret` that creates a Secret, references it from an Integration, and verifies the resolved value reaches the Coralogix backend. Mirrors the existing `Connector with SecretKeyRef` test.
- `go build ./...` and `go vet` clean.
- CRDs regenerated with controller-gen v0.17.2; `docs/api.md` regenerated with `crdoc`.

## Sample

`config/samples/v1alpha1/integrations/gcp-metrics-collector-with-secret.yaml` shows the full pattern: a `Secret` holding the SA key plus an `Integration` referencing it.

## Out of scope

- Workload Identity Federation / short-lived tokens — separate concern, requires upstream API changes.
- ConfigMap-backed parameters — easy to add later if a use case appears, but no current integration needs it.
- Watching referenced Secrets to trigger re-reconcile on rotation — relies on the controller's existing requeue interval for now; can be added in a follow-up if rotation latency becomes an issue.


assafad1:
@codex review

api/coralogix/v1alpha1/integration_types.go
```diff
@@ -36,46 +38,81 @@ type IntegrationSpec struct {
 	// +kubebuilder:pruning:PreserveUnknownFields
 	// Parameters required by the integration.
 	Parameters runtime.RawExtension `json:"parameters"`
```
assafad1: should we make this optional, so you can set only sensitive params?

assafad1 (approved):

PR #524, opened 2026-06-29
AI Custom Evaluations

markshleifer-coralogix:
Implemented AI Custom Evaluation

Manually tested:
- Creation of a resource
- Linking and unlinking applications
- Changing in the fields: `policyType`, `description`, `instructions`, `shouldIncludeSystemPrompt`, `flags` and `examples`.
- Deletion of a resource


assafad1:
@codex review

api/coralogix/v1alpha1/aicustomevaluation_types.go
```diff
@@ -0,0 +1,285 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1alpha1
+
+import (
+	"fmt"
+
+	aievaluations "github.com/coralogix/coralogix-management-sdk/go/openapi/gen/ai_evaluations_service"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
assafad1: Nit: follow other files imports grouping

api/coralogix/v1alpha1/aicustomevaluation_types.go
```diff
@@ -0,0 +1,285 @@
...
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=256
+	// +kubebuilder:validation:Enum=quality;security
+	PolicyType string `json:"policyType"`
+
+	// Human-readable description. Defaults to an empty string.
+	// +optional
+	// +kubebuilder:default=""
+	// +kubebuilder:validation:MaxLength=65536
+	Description string `json:"description,omitempty"`
```
assafad1: Why not using pointers for optional fields?

api/coralogix/v1alpha1/aicustomevaluation_types.go
```diff
@@ -0,0 +1,285 @@
...
+	return examples, acceptable.Flags, prohibited.Flags, nil
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.printableStatus"
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
+
+// AICustomEvaluation is the Schema for the AI custom evaluations API.
+type AICustomEvaluation struct {
```
assafad1, line 241: Can you link to Coralogix docs? like other crds

assafad1 (commented):
Please add working examples to https://github.com/coralogix/coralogix-operator/tree/main/config/samples/v1alpha1. I think AI evaluation CRD samples are also missing.

api/coralogix/v1alpha1/aicustomevaluation_types.go
```diff
@@ -0,0 +1,285 @@
...
+	// Resolved AI application IDs linked to this custom evaluation.
+	// +optional
+	// +listType=set
+	ApplicationIds []string `json:"applicationIds,omitempty"`
+
+	// Resolved AI application mappings linked to this custom evaluation.
+	// +optional
+	// +listType=map
+	// +listMapKey=id
+	Applications []AICustomEvaluationApplicationStatus `json:"applications,omitempty"`
```
assafad1: Can you please explain why do we need these fields in the status? It will be the first CRD that adds custom fields to the status.

markshleifer-coralogix: Good point. After I addressed the comment by codex, these fields are no longer needed for reconciliation.

The operator now resolves spec.applications to IDs during reconcile, so the storing the links of application name -> application IDs is not needed anymore.

markshleifer-coralogix:
@codex review

api/coralogix/v1alpha1/aicustomevaluation_types.go
```diff
@@ -0,0 +1,269 @@
...
+
+	aievaluations "github.com/coralogix/coralogix-management-sdk/go/openapi/gen/ai_evaluations_service"
+)
+
+const (
+	AICustomEvaluationPolicyTypeQuality  = "quality"
+	AICustomEvaluationPolicyTypeSecurity = "security"
+
+	aiCustomEvaluationAcceptableScore    = "1"
+	aiCustomEvaluationProhibitedScore    = "0"
```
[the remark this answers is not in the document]
markshleifer-coralogix: Made sure with the BE team, this is correct:
```
aiCustomEvaluationAcceptableScore    = "1"
aiCustomEvaluationProhibitedScore    = "0"
```

The BE will update the examples on their side so they will be clearer.

api/coralogix/v1alpha1/aicustomevaluation_types.go
```diff
@@ -0,0 +1,269 @@
...
+			acceptable = *s.Criteria.Acceptable
+		}
+		if s.Criteria.Prohibited != nil {
+			prohibited = *s.Criteria.Prohibited
+		}
+	}
+
+	if len(acceptable.Examples)+len(prohibited.Examples) > 100 {
+		return nil, "", "", fmt.Errorf("criteria can include at most 100 total examples across acceptable and prohibited criteria")
+	}
```
assafad1: What's that for? Is it a backend limitation?
If so, can we validate on CRD schema so that cluster rejects it (that would improve UX - faster feedback)?

api/coralogix/v1alpha1/aicustomevaluation_types.go
```diff
@@ -0,0 +1,269 @@
...
+		return false
+	}
+	return *s.ShouldIncludeSystemPrompt
+}
+
+func (s *AICustomEvaluationSpec) DescriptionValue() string {
+	return ptr.Deref(s.Description, "")
+}
+
+func (s *AICustomEvaluationSpec) ExtractAICustomEvaluationCriteria() ([]aievaluations.CustomEvaluationExample, string, string, error) {
```
assafad1: It feels like this function should be split to 3, or am I missing something?

internal/controller/coralogix/v1alpha1/aicustomevaluation_controller.go
```diff
@@ -0,0 +1,340 @@
...
+			ids := append([]string(nil), customEvaluation.GetApplicationIds()...)
+			sort.Strings(ids)
+			return ids, nil
+		}
+	}
+
+	return nil, fmt.Errorf("remote AICustomEvaluation %q not found", customEvaluationID)
+}
+
+func (r *AICustomEvaluationReconciler) reconcileApplicationLinks(ctx context.Context, customEvaluationID string, currentApplicationIDs []string, desiredApplicationIDs []string) error {
```
assafad1, line 264: This is a complication that should be part of the API, as discussed, no?

internal/controller/coralogix/v1alpha1/aicustomevaluation_controller.go
```diff
@@ -0,0 +1,340 @@
...
+		}
+		log.Info("Remote AICustomEvaluation examples cleared", "response", utils.FormatJSON(examplesUpdateResponse))
+	}
+
+	remoteApplicationIDs, err := r.getRemoteAICustomEvaluationApplicationIDs(ctx, id)
+	if err != nil {
+		return err
+	}
+
+	desiredApplicationIDs := aiCustomEvaluationApplicationIDs(applications)
```
assafad1, line 152: Also I wonder why on creation this isn't needed

assafad1 (commented):
@markshleifer-coralogix I think this implementation includes many workarounds that we agreed should be part of the API. Correct me if I'm wrong.

api/coralogix/v1alpha1/aicustomevaluation_types.go
markshleifer-coralogix: Remove from here. It's only in the schema now.

internal/controller/coralogix/v1alpha1/aicustomevaluation_controller.go
markshleifer-coralogix: `CreateCustomEvaluationRequest` has an `application_ids` field, so the server links everything atomically as part of create. `UpdateCustomEvaluationRequest` has no equivalent field — only single-application `LinkCustomEvaluation`/`UnlinkCustomEvaluationFromApp` endpoints exist. So on update the operator has to fetch the current links, diff against desired, and call link/unlink one at a time to get the same end state.

internal/controller/coralogix/v1alpha1/aicustomevaluation_controller.go
markshleifer-coralogix: See the comment above: https://github.com/coralogix/coralogix-operator/pull/524#discussion_r3644976146

internal/controller/coralogix/v1alpha1/aicustomevaluation_controller.go
markshleifer-coralogix: As we agreed to continue with the ticket since the TF part was already merged, I would add that it's possible to address the API gap in a backward compatible way later: the create and update requests can accept application names directly (matching how the sibling `AIEvaluation` API already takes application/subsystem by name). This would allow us to remove the `resolveApplications` and `reconcileApplicationLinks` parts completely.

api/coralogix/v1alpha1/aicustomevaluation_types.go
```diff
@@ -0,0 +1,257 @@
...
+	// +optional
+	Prohibited *AICustomEvaluationCriterion `json:"prohibited,omitempty"`
+}
+
+type AICustomEvaluationCriterion struct {
+	// Criterion flags.
+	// +optional
+	// +kubebuilder:default=""
+	// +kubebuilder:validation:MaxLength=65536
+	Flags string `json:"flags,omitempty"`
```
assafad1: Does it need omitempty if we have `""` default? Can you please make sure it's right for the rest of the fields as well?

assafad1 (dismissed):

api/coralogix/v1alpha1/aicustomevaluation_types.go
markshleifer-coralogix: Removed the "" default and made it a string pointer since it's optional.

assafad1 (approved):

PR #537, opened 2026-07-21
Added an option to import dashboards

markshleifer-coralogix:
# Add support for importing existing dashboards

Adds the ability to adopt a pre-existing remote Coralogix dashboard into a `Dashboard` CR, instead of only being able to create new ones.

## Example

```yaml
apiVersion: coralogix.com/v1alpha1
kind: Dashboard
metadata:
  name: imported-dashboard
  annotations:
    app.coralogix.com/import-id: "3d7f1c2a-9e4b-4a11-8f2d-1a2b3c4d5e6f"
spec:
  json: |
    { ... }   # required; will overwrite the remote dashboard's content on the next reconcile
```

On the first reconcile, if `app.coralogix.com/import-id` is set and `status.id` is empty, the controller fetches the existing dashboard by ID instead of creating a new one, and sets `status.id` accordingly. From then on the CR behaves like any other Dashboard CR: `spec` is the source of truth and will overwrite the remote dashboard's content on the next reconcile if it differs.

Once adoption succeeds, `status.imported` is set to `true` and persists for the lifetime of the CR. Unlike `status.id`, it is never cleared — even if the remote dashboard is later deleted outside the operator — so a subsequent reconcile recreates the dashboard from `spec` instead of retrying the import lookup for an ID that no longer exists.

Embedding an `id` field inside `spec.json`/`spec.gzipJson`/the referenced ConfigMap is now explicitly rejected with a clear error, for both create and update — there is no legitimate use for it once import is annotation-driven.

## Why an annotation instead of a spec field

The import ID is only meaningful once, on the very first reconcile before `status.id` is set. A spec field would remain part of the schema forever but go inert after adoption, which breaks the general expectation that editing `spec` affects reconciliation. An annotation doesn't carry that expectation, and follows the existing `app.coralogix.com/<key>` annotation convention already used in the operator (e.g. `LogVerbosityAnnotationKey`). This also mirrors precedent elsewhere (Crossplane's `crossplane.io/external-name`, AWS ACK's adoption annotations).

## Tests

- Added e2e test coverage for the import flow.
- Manually verified:
  - Import an existing dashboard by ID, then update its spec — update is applied correctly.
  - Importing with an `id` present in `spec` fails, as expected.
  - Importing with an unknown/nonexistent ID fails, as expected (no fallback to creating a duplicate).
  - Deleting an imported dashboards in the UI recreates it.
  - Updating a CR with an `id` present in `spec` fails, as expected.

### DIFF
--- a/internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
+++ b/internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
@@ -125,8 +125,12 @@ func ReconcileResource(ctx context.Context, req ctrl.Request, obj coralogix.Obje
 			return ManageErrorWithRequeue(ctx, obj, utils.ReasonInternalK8sError, err)
 		}
 
-		if err := removeField(ctx, obj, "status"); err != nil {
-			log.Error(err, "Error removing id from status")
+		var preserve []string
+		if p, ok := obj.(statusFieldPreserver); ok {
+			preserve = p.PreservedStatusFields()
+		}
+		if err := clearStatusExcept(ctx, obj, preserve...); err != nil {
+			log.Error(err, "Error clearing status")
 			return ManageErrorWithRequeue(ctx, obj, utils.ReasonInternalK8sError, err)
 		}
 
@@ -165,6 +169,44 @@ func removeField(ctx context.Context, obj client.Object, fields ...string) error
 	}
 
 	unstructured.RemoveNestedField(u.Object, fields...)
+
+	return config.GetClient().Status().Update(ctx, u)
+}
+
+// statusFieldPreserver is implemented by resource types that need one or more of their
+// status fields to survive clearStatusExcept, e.g. Dashboard's status.imported, which
+// must persist across a remote deletion so a subsequent adoption recreates instead of
+// re-attempting an import against an id that no longer exists.
+type statusFieldPreserver interface {
+	PreservedStatusFields() []string
+}
+
+// clearStatusExcept resets obj's status to only the named top-level fields, discarding
+// everything else (id, conditions, printableStatus, etc.).
+func clearStatusExcept(ctx context.Context, obj client.Object, keep ...string) error {
+	u := &unstructured.Unstructured{}
+	if err := config.GetScheme().Convert(obj, u, nil); err != nil {
+		return fmt.Errorf("failed to convert object to unstructured: %w", err)
+	}
+
+	status, found, err := unstructured.NestedMap(u.Object, "status")
+	if err != nil {
+		return fmt.Errorf("failed to read status: %w", err)
+	}
+	if !found {
+		return nil
+	}
+
+	keepSet := make(map[string]bool, len(keep))
+	for _, field := range keep {
+		keepSet[field] = true
+	}
+
+	for field := range status {
+		if !keepSet[field] {
+			unstructured.RemoveNestedField(u.Object, "status", field)
+		}
+	}
 
 	return config.GetClient().Status().Update(ctx, u)
 }
--- a/internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
+++ b/internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
@@ -125,11 +125,7 @@ func ReconcileResource(ctx context.Context, req ctrl.Request, obj coralogix.Obje
 			return ManageErrorWithRequeue(ctx, obj, utils.ReasonInternalK8sError, err)
 		}
 
-		var preserve []string
-		if p, ok := obj.(statusFieldPreserver); ok {
-			preserve = p.PreservedStatusFields()
-		}
-		if err := clearStatusExcept(ctx, obj, preserve...); err != nil {
+		if err := clearRemoteStatus(ctx, obj); err != nil {
 			log.Error(err, "Error clearing status")
 			return ManageErrorWithRequeue(ctx, obj, utils.ReasonInternalK8sError, err)
 		}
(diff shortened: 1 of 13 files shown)


markshleifer-coralogix:
@codex review

internal/controller/coralogix/v1alpha1/dashboard_controller.go
```diff
@@ -64,6 +65,22 @@ func (r *DashboardReconciler) HandleCreation(ctx context.Context, log logr.Logge
...
+	if importID := dashboard.GetAnnotations()[internalutils.ImportDashboardIDAnnotationKey]; importID != "" {
+		log.Info("Import annotation present, adopting existing remote dashboard", "id", importID)
+		getResponse, err := r.DashboardsClient.Get(ctx, &cxsdk.GetDashboardRequest{DashboardId: wrapperspb.String(importID)})
+		if err != nil {
+			return fmt.Errorf("error on getting remote dashboard %q for import: %w", importID, err)
+		}
+		dashboard.Status = coralogixv1alpha1.DashboardStatus{
+			ID: ptr.To(getResponse.Dashboard.GetId().GetValue()),
+		}
+		return nil
```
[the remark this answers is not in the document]
markshleifer-coralogix: If the update fails later, it should be the same as a regular existing dashboard with a failed update, and it's the behavior now.

internal/controller/coralogix/v1alpha1/dashboard_controller.go
```diff
@@ -64,6 +65,22 @@ func (r *DashboardReconciler) HandleCreation(ctx context.Context, log logr.Logge
 	if err != nil {
 		return fmt.Errorf("error on extracting dashboard from spec: %w", err)
 	}
+	if err = validateNoEmbeddedID(dashboardToCreate); err != nil {
+		return err
+	}
+
+	if importID := dashboard.GetAnnotations()[internalutils.ImportDashboardIDAnnotationKey]; importID != "" {
```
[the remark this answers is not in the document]
markshleifer-coralogix: Fixed by adding a status field that will indicate that an import has happened.
The new flow will run only when an import hasn't happened yet.

internal/utils/consts.go
```diff
@@ -56,4 +56,14 @@ const (
...
+
+	// ImportDashboardIDAnnotationKey adopts a pre-existing remote Dashboard into a Dashboard CR by ID.
+	// Once the dashboard is adopted it is treated exactly like a normally-created Dashboard CR,
+	// and the CR's spec becomes the source of truth and overwrites the remote dashboard's content
+	// on the next reconcile. The annotation itself is not removed after adoption (mirroring
+	// adoption annotations in tools like Crossplane and AWS ACK); instead status.imported on the
+	// Dashboard tracks that adoption already happened, so that if the remote dashboard is later
+	// deleted outside the operator, it is recreated from spec instead of the import path
+	// retrying a Get for an ID that no longer exists.
+	ImportDashboardIDAnnotationKey = "app.coralogix.com/import-id"
```
assafad1: nit: move the const to dashboards file, or generalize const name

internal/controller/coralogix/v1alpha1/dashboard_controller.go
```diff
@@ -75,18 +101,32 @@ func (r *DashboardReconciler) HandleCreation(ctx context.Context, log logr.Logge
...
+
 func (r *DashboardReconciler) HandleUpdate(ctx context.Context, log logr.Logger, obj client.Object) error {
 	dashboard := obj.(*coralogixv1alpha1.Dashboard)
 	dashboardToUpdate, err := dashboard.Spec.ExtractDashboardFromSpec(ctx, dashboard.Namespace)
 	if err != nil {
 		return fmt.Errorf("error on extracting dashboard from spec: %w", err)
 	}
+	if err = validateNoEmbeddedID(dashboardToUpdate); err != nil {
+		return err
+	}
```
assafad1: wouldn't it break every existing CR with id in spec?

tests/e2e/dashboard_test.go
```diff
@@ -109,6 +110,70 @@ var _ = Describe("Dashboard", Ordered, func() {
...
+		fetchedDashboard := &coralogixv1alpha1.Dashboard{}
+		Eventually(func(g Gomega) error {
+			g.Expect(crClient.Get(ctx, types.NamespacedName{Name: dashboardName, Namespace: testNamespace}, fetchedDashboard)).To(Succeed())
+			g.Expect(meta.IsStatusConditionTrue(fetchedDashboard.Status.Conditions, utils.ConditionTypeRemoteSynced)).To(BeTrue())
+			if fetchedDashboard.Status.ID != nil {
+				return nil
+			}
+			return fmt.Errorf("Dashboard ID is not set")
+		}, time.Minute, time.Second).Should(Succeed())
+		Expect(*fetchedDashboard.Status.ID).To(Equal(dashboardID))
```
assafad1, line 162: maybe check `Status.Imported` is true? :)

assafad1 (changes requested):
I'm trying to figure out if this change introduces a bug with selectors here: https://github.com/coralogix/coralogix-operator/blob/fa916e252f8e55a690071bd5c38b7128f4d94b82/internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go#L128

Let's think of the following flow (probably edge case, but still):
- Operator is configured to operate on `env: prod` labeled CRs
- User creates a dashboard CR with a `env: prod` label and an import annotation
- Operator imports the dashboard as expected
- User removes/changes the label in CR, so the operator deletes the remote resource, and remove the whole status (including `imported` field).
- User labels it `env: prod` again
- Operator operates on it again, sees the import annotation, but the remote resource was deleted so it will constantly fail importing (on all future reconciles).

I think if we keep `imported` true when this happens, the operator will recreate this dashboard as expected. WDYT?

internal/controller/coralogix/v1alpha1/dashboard_controller.go
markshleifer-coralogix: Using an id when creating will not work, therefore the only way to have an id in the spec is to have it added later (which is ignored anyway). I think it still makes sense to return an error here, but I will narrow it down to the scenario of when the import annotation exists. WDYT?

assafad1: > Using an id when creating will not work

Even when the ID isn't used by an existing dashboard?

markshleifer-coralogix: No, only for imported ids that already exist. My statement above was misleading:
Creating it works, however changing it later is ignored. It uses the old id.

However I would still keep this validation in case of an import validation. Keeping both ids (spec + import) together can be confusing.

markshleifer-coralogix:
@assafad1 Addressed your comments.
The issue you have mentioned was a regression. Fixed it and added a test (verified locally as well). The status will be removed except for imported, as you suggested.
Regarding `validateNoEmbeddedID`, it's narrower now: it will validate only when importing. If you find it too strict as well, let's discuss.

internal/controller/coralogix/coralogix-reconciler/coralogix_reconciler.go
```diff
@@ -125,8 +125,12 @@ func ReconcileResource(ctx context.Context, req ctrl.Request, obj coralogix.Obje
...
 		}
 
-		if err := removeField(ctx, obj, "status"); err != nil {
-			log.Error(err, "Error removing id from status")
+		var preserve []string
+		if p, ok := obj.(statusFieldPreserver); ok {
+			preserve = p.PreservedStatusFields()
+		}
+		if err := clearStatusExcept(ctx, obj, preserve...); err != nil {
+			log.Error(err, "Error clearing status")
```
assafad1: Can we instead use existing `removeField()` for every field we need to remove? just to keep things simple.
i.e. 

```suggestion
        removeField(ctx, obj, "status", "id")
        removeField(ctx, obj, "status", "conditions")
        removeField(ctx, obj, "status", "printableStatus")
```
or am I missing something here? :)

assafad1: Then we can remove [this part](https://github.com/coralogix/coralogix-operator/pull/537/changes#diff-5ccf3e216730d16305b2c4124a3cc7a0a061bf4024eb33e0a62ba5d3b9927eddR176-R213)

markshleifer-coralogix: It will work, but then we won't be forward compatible, no?
If a new custom field would be added, we might forget to remove it here.

assafad1: But with current implementation, if you want to preserve a new field (e.g. `imported` for a new CRD), you'd have to remember expanding this func (if it's for dashboards, or create a new method if other CRD). 
```
func (d *Dashboard) PreservedStatusFields() []string {
	return []string{"imported"}
}
```
If we forget, this regression I found will be re-introduced, right? And I would say it's more likely to happen next, than a new removable field (like `id`). Plus the code is longer and more complicated. 

I'd say we lack some testing here, but it shouldn't be a part of this task.

markshleifer-coralogix: Addressed it, see the new commit.
I did not use the suggested implementation since calling `removeField` multiple times will revive the removed fields of the previous attempt (the field is removed in a local copy `u` in removeField, so the original obj still has the removed filed):
```
func removeField(ctx context.Context, obj client.Object, fields ...string) error {
	u := &unstructured.Unstructured{}
	if err := config.GetScheme().Convert(obj, u, nil); err != nil {
		return fmt.Errorf("failed to convert object to unstructured: %w", err)
	}

	unstructured.RemoveNestedField(u.Object, fields...)

	return config.GetClient().Status().Update(ctx, u)
}
```

assafad1 (approved):

PR #546, opened 2026-07-31
Migrate events2metrics from gRPC to rest client

markshleifer-coralogix:
## Summary

Migrate the `Events2Metric` controller from the gRPC SDK client to the OpenAPI REST client. Conversion now builds OpenAPI request models instead of protobuf wrappers. The CRD schema is unchanged.

## Backward compatibility

- Existing `Events2Metric` manifests keep the same fields and values. No CRD breaking change.
- Create/update/delete still go through the shared reconciler. Status `id` and `RemoteSynced` behavior are unchanged.
- Replace presence semantics are unchanged for users:
  - Omitting `description` still clears the remote value. REST omit would keep the old value, so the operator sends `""` on replace (implementation detail only).
  - Omitting `lucene` / `dataSource` / `alias` still clears them remotely.
  - Omitting `permutationsLimit` still omits it on replace; the API still resets the limit to its default (`30000`).

## E2E coverage

- Logs path: create, update (name, description, labels, aggregations), clear optional fields, delete; asserts remote state over REST.
- Spans path: create/update/clear/delete with `samples` and `histogram` aggregations and bucket metadata.
- Unique resource/metric names to avoid account-wide collisions; cleanup via `DeferCleanup`.
- Unit tests cover create/replace request wire shape for logs and spans.

### DIFF
--- a/api/coralogix/utils.go
+++ b/api/coralogix/utils.go
@@ -17,7 +17,6 @@ package coralogix
 import (
 	"fmt"
 
-	"google.golang.org/protobuf/types/known/wrapperspb"
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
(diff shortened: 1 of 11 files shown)


markshleifer-coralogix:
@codex review

povilasv (approved):

PR #548, opened 2026-08-05
refactor: drop the operator's gRPC dependency

assafad1:
## Why

BYOC + FedRAMP tenants behind PrivateLink only have `api.<domain>` published — `ng-api-grpc.<domain>` and `ng-api-http.<domain>` can't be created. The operator derived both from the same `--region`/`--domain` value, so anything on those hosts was unreachable.

After this PR the operator talks to **`api.<domain>` only**.

## Changes

1. **ArchiveLogsTarget → REST.** The last real gRPC call (`cxsdk.ArchiveRetentionsClient.Activate`) now goes through `oapiClientSet.ArchiveRetentions().RetentionsServiceActivateRetentions(...)` — the same REST client TCOLogsPolicies/TCOTracesPolicies already use.
2. **Group → narrow client.** `ClientSet.Users()` was never gRPC (SCIM over HTTP); the controller just held the whole gRPC `ClientSet` to reach it. It now takes a `*cxsdk.UsersClient`, so **no gRPC ClientSet is constructed anywhere** and nobody can reintroduce a gRPC call through it.
3. **SDK bump → coralogix-management-sdk#666** (merged, `a572d270`), which serves SCIM from `api.<domain>` instead of `ng-api-http.<domain>`. This is what removes the last non-`api` host.
4. **Prepare 2.5.0** — `OperatorVersion` bumped to `2.5.0` in `cmd/main.go` (chart intentionally untouched).
5. **Cleanups:** `Config.CoralogixGrpcUrl` → `CoralogixRegionOrDomain` (no user-facing flag/env change — `--region`/`--domain`/`CORALOGIX_REGION`/`CORALOGIX_DOMAIN` all unchanged).
6. ⚠️ **Behavior change:** `operator_info`'s `coralogix_url` label reported the derived gRPC endpoint, which the operator no longer contacts. It now reports `CoralogixOpenApiUrl`. Anything keying on that label value will see it change.

## Verification

`make unit-tests` and `make lint` (0 issues) green. Re-run live against **eu2** on a dedicated kind cluster at this exact commit — 29 controllers started, **zero errors across the whole run**:

- **Archive retentions (the migrated call):** ArchiveLogsTarget create → delete; tenant `isActive` false → true → false, `RemoteSynced`, and zero `error activating archive retentions`. Debug log shows `Creating remote archivelogstarget` → `Remote archivelogstarget created`; the activate call sits between them and returns before status is set, so success implies 2xx.
- **Users / SCIM (now over `api.<domain>`):** Group with `.spec.members` created, updated and deleted. The member resolved on the tenant to `userId 57108c2e-…`, identical to what `api.eu2.coralogix.com/scim/Users` returns for that username.
- **SDK bump is effective:** constructing `NewUsersClient` from the pinned module version (resolved from the proxy, no local replace) yields `https://api.<domain>/scim/Users` for regions, custom domains, and `api.`-prefixed domains alike — re-checked against the merged `a572d270`, whose content is byte-identical to the commit the live run used.
- **Release prep:** `cx_operator_build_info{operator_version="2.5.0", coralogix_url="https://api.eu2.coralogix.com/mgmt/openapi/5"}` — confirms both the version bump and the label change.

Test resources (group, role, scope, archive target) were removed afterwards and verified gone (404s).

### DIFF
--- a/api/coralogix/v1alpha1/group_types.go
+++ b/api/coralogix/v1alpha1/group_types.go
@@ -94,13 +94,13 @@ type ResourceRef struct {
 
 func (g *Group) ExtractCreateGroupRequest(
 	ctx context.Context,
-	cxClient *cxsdk.ClientSet) (*groups.CreateTeamGroupRequest, error) {
+	usersClient *cxsdk.UsersClient) (*groups.CreateTeamGroupRequest, error) {
 	var groupType *groups.GroupType
 	if g.Spec.GroupType != nil {
 		groupType = groupTypeSchemaToOpenAPI[*g.Spec.GroupType].Ptr()
 	}
 
-	usersIds, err := g.ExtractUsersIDs(ctx, cxClient)
+	usersIds, err := g.ExtractUsersIDs(ctx, usersClient)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


assafad1:
@codex review

assafad1:
@codex review

markshleifer-coralogix (approved):

PR #549, opened 2026-08-06
fix: make `priority` required on TCO policy targets

assafad1:
## What

Makes `priority` required on every `TCOLogsPolicies` target.

Follow-up to #545 (CX-51353), which modelled `TCOPolicyTarget.Priority` as optional (`*string`, `+optional`) on the assumption that a priority-less target would inherit the policy-level priority. Based on our feedback, the backend has changed and now rejects that shape.

**Not a breaking change** — targets support has not been released yet.

## Why — live API verification

Probed against EU2 (`PUT /dataplans/log-policies/v1`). Every probe was cleaned up afterwards; the tenant was restored to `{"policies":[]}`.

| Request | Result |
|---|---|
| policy `priority: MEDIUM` + target with no `priority` | **400** `Target "logs": a priority is required on every target.` |
| target `priority: PRIORITY_TYPE_UNSPECIFIED` | **400** same error — `UNSPECIFIED` counts as unset |
| target with only `priorityOverride`, no `priority` | **400** same error — an override does **not** satisfy it |
| 2 targets, one missing `priority` | **400** `Target "logs2": a priority is required on every target.` |
| policy priority omitted **or** `UNSPECIFIED` + every target priced | **200** ✅ |
| policy `priority` set, no targets | **200** ✅ |
| no priority anywhere | **400** `A priority is required: set the policy priority (or a quota-based priority override), or provide targets that each specify a priority.` |
| target with both `priority` and `priorityOverride` | **200** ✅ — these two are *not* mutually exclusive at target level |

## Changes

- `TCOPolicyTarget.Priority`: `*string` + `+optional` → required `string`.
- `expandTCOPolicyTargets` maps the priority unconditionally instead of behind a nil guard, matching how the already-required `TCOPolicyUsageTier.Priority` is handled.
- Policy-level `Priority` doc comment corrected — it no longer claims targets can inherit it.
- Regenerated: DeepCopy, `config/crd/bases`, `docs/api.md`, and the Helm chart CRD (synced by hand; `diff` against the base shows only the two `{{- if .Values.crds.create }}` guard lines). `priority` now joins `dataset` in `targets[].required`.
- The unit test asserting *"policy priority + priority-less targets is accepted"* becomes a rejection test — that combination is now always a 400, per the mutual-exclusion rule above.
- `config/samples/.../example-tco-logs-policies.yaml` needed no change; it already prices every target.

## Scope note

The backend enforces a **second** rule: policy-level and per-target priority are mutually exclusive. This PR deliberately does **not** add CEL admission validation for it — violations reach the API and surface as `RemoteSynced=False` with the backend message. Happy to add the guard in a follow-up if we'd rather fail at admission time.

## Testing

- `make unit-tests` — all packages pass. The v1alpha1 envtest suite ran 49/49; verified the new rejection spec actually executes rather than passing vacuously.
- `make lint` — 0 issues. `go vet ./api/... ./internal/... ./tests/...` clean.
- Live end-to-end against EU2: replayed the full sample YAML exactly as the updated `extractCreateLogPolicyRequest` serialises it (all four policies, archive-retention IDs resolved). HTTP 200, and `fourth policy with targets` came back with `myDataset/LOW` and `archiveDataset/LOW` + quota override. Tenant cleaned up.

### DIFF
--- a/api/coralogix/v1alpha1/tcologspolicies_types.go
+++ b/api/coralogix/v1alpha1/tcologspolicies_types.go
@@ -45,7 +45,7 @@ type TCOLogsPolicy struct {
 	Description *string `json:"description,omitempty"`
 
 	// +kubebuilder:validation:Enum=block;high;medium;low
-	// The policy priority. Required when targets is not set, or when targets do not specify their own priorities.
+	// The policy priority. Required when targets is not set. Mutually exclusive with targets, which carry their own per-target priorities.
 	// +optional
 	Priority *string `json:"priority,omitempty"`
 
@@ -126,9 +126,8 @@ type TCOPolicyTarget struct {
 	Dataspace *string `json:"dataspace,omitempty"`
 
 	// +kubebuilder:validation:Enum=block;high;low;medium
-	// +optional
-	// Per-target priority. Mutually exclusive with a policy-level priority.
-	Priority *string `json:"priority,omitempty"`
+	// Per-target priority. Required on every target. Mutually exclusive with the policy-level priority.
+	Priority string `json:"priority"`
 
 	// +optional
 	// Dynamic quota-based priority override for this target.
@@ -303,14 +302,12 @@ func expandTCOPolicyTargets(retentionsByName map[string]string, targets []TCOPol
 			errs = errors.Join(errs, err)
 			continue
 		}
+		priority := PrioritySchemaToOpenAPI[target.Priority]
 		t := tcopolicies.V1Target{
 			Dataset:          &target.Dataset,
 			Dataspace:        target.Dataspace,
+			Priority:         &priority,
 			ArchiveRetention: archiveRetention,
-		}
-		if target.Priority != nil {
-			priority := PrioritySchemaToOpenAPI[*target.Priority]
-			t.Priority = &priority
 		}
 		if target.PriorityOverride != nil {
 			t.PriorityOverride = expandTCOPolicyPriorityOverride(target.PriorityOverride)
--- a/api/coralogix/v1alpha1/tcologspolicies_types.go
+++ b/api/coralogix/v1alpha1/tcologspolicies_types.go
@@ -68,8 +68,8 @@ type TCOLogsPolicy struct {
 	// +optional
 	Subsystems *TCOPolicyRule `json:"subsystems,omitempty"`
 
-	// Targets defines the datasets to route matched data to, each with its own priority.
-	// When set, overrides or supplements the policy-level priority.
+	// Targets defines the datasets to route matched data to, each with its own required priority.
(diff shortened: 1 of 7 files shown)


assafad1:
@codex review

api/coralogix/v1alpha1/tcologspolicies_types.go
```diff
@@ -45,7 +45,7 @@ type TCOLogsPolicy struct {
 	Description *string `json:"description,omitempty"`
 
 	// +kubebuilder:validation:Enum=block;high;medium;low
-	// The policy priority. Required when targets is not set, or when targets do not specify their own priorities.
+	// The policy priority. Required when targets is not set. Mutually exclusive with targets, which carry their own per-target priorities.
```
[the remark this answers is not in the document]
assafad1: Good catch — fixed in 1b96010.

The `Targets` comment said targets "overrides or supplements the policy-level priority", which is exactly the shape the backend rejects:

> `priority/priorityOverride must be set either at the policy level or per-target, not both.`

Reworded to:

```go
// Targets defines the datasets to route matched data to, each with its own required priority.
// Setting targets requires omitting the policy-level priority; the two are mutually exclusive.
```

Regenerated so the new text propagates to `config/crd/bases`, the Helm chart CRD, and `docs/api.md`.

Note the mutual exclusion is still enforced only by the backend — this PR deliberately does not add a CEL guard, so setting both passes admission and surfaces as `RemoteSynced=False`. The comment now at least tells users not to do it. Happy to add the CEL rule in a follow-up if we prefer failing at admission time.

api/coralogix/v1alpha1/tcologspolicies_types.go
```diff
@@ -126,9 +126,8 @@ type TCOPolicyTarget struct {
 	Dataspace *string `json:"dataspace,omitempty"`
 
 	// +kubebuilder:validation:Enum=block;high;low;medium
```
markshleifer-coralogix, line 128: It's more related to https://github.com/coralogix/coralogix-operator/pull/545, but I was wondering, do you think it makes sense to limit this to `low` and `medium` only?

From: https://coralogix.com/docs/user-guides/data-layer/default-dataspace/user-defined-datasets/
Image

assafad1: I don't think so. A user can set `default/logs` with `HIGH` and `BLOCK`.

markshleifer-coralogix: Got it, I thought that targets are only for user-defined datasets.

markshleifer-coralogix (approved):

PR #551, opened 2026-08-06
fix: restrict GITHUB_TOKEN permissions in workflows (11 code scanning alerts) + bump Go to 1.26

assafad1:
Closes all 11 open code-scanning alerts, which were the same rule — `actions/missing-workflow-permissions`. None of these 10 workflows declared a `permissions` block, so every job ran with the default read/**write** `GITHUB_TOKEN`, including test-only jobs.

Most get `contents: read`. Two genuinely need write, verified against real runs:

- **`unit-tests`** — `go-patch-cover-action` pushes `refs/notes/coverage` on push and comments on PRs. One job serves both events, and permissions can't be event-conditional, so it takes the union.
- **`helm-update`** — `create-pull-request` pushes a branch and opens the PR with `secrets.GITHUB_TOKEN`.

Second commit bumps **Go 1.25 → 1.26** (`go.mod` + `Dockerfile`). 

`actionlint` reports the same findings before and after; `make unit-tests` and lint pass on go1.26.3 with no generated-file drift.

One thing to check before un-drafting: **the go-patch-cover comment should still appear on this PR** — that's the live test of `pull-requests: write`. The push path only runs after merge.

### DIFF
--- a/.github/actions/build-operator-image/action.yml
+++ b/.github/actions/build-operator-image/action.yml
@@ -14,8 +14,8 @@ runs:
   using: composite
   steps:
     # The released image builds inside the `golang:` tag in Dockerfile, so read the version from
-    # there rather than from go.mod. go.mod's `go` directive is the module's *minimum* (1.24.0),
-    # which is not what ships: `golang:1.24` resolves to the newest 1.24.x. Taking go.mod at face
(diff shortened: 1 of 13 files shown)


assafad1:
@codex review

markshleifer-coralogix (approved):

PR #554, opened 2026-08-18
fix: stop Group and Scope e2e specs from colliding on the shared CI account

cazorla19:
## Summary
- Upgrade-test jobs share one Coralogix account. Group e2e used a fixed `scope-for-group` name and second-granularity CustomRole/Group names, so parallel matrix jobs (and other e2e workflows) could create the same backend role/scope. One job then deleted it; the other Group reconcile 403'd with `Permission Denied: Trying to perform operation on roles which are not of type team or do not belong to this team` and never reached `RemoteSynced`.
- Give Group, Scope, and CustomRole e2e names nanosecond uniqueness (including update names like `custom-role-updated`). Wait for Scope and CustomRole to be RemoteSynced before creating the Group, matching the View/ViewFolder pattern.

## Test plan
- [ ] Confirm `go test -c ./tests/e2e/` still compiles
- [ ] Watch upgrade-test / e2e-tests on this PR: Group and Scope specs should pass even when matrix jobs overlap
- [ ] Teams live create remains skipped unless local `MS_TEAMS_*` env vars are set (not part of this PR)

Made with [Cursor](https://cursor.com)


cazorla19:
Follow-up for the remaining upgrade-test (0.5) failure: `ArchiveMetricsTarget Should be set successfully`.

That spec already reached `RemoteSynced`, then waited 3 minutes for `TenantConfig.Disabled == false`. Archive metrics is an account-wide singleton. A sibling job (upgrade 1.0 / e2e / helm-e2e) deleted its CR and called `DisableArchive`. This operator’s periodic `HandleUpdate` only sent `UpdateTenant` (retention + S3), which does not re-enable the archive — unlike ArchiveLogs, which already calls `SetTarget(active=true)` on update.

`HandleUpdate` now configure-then-updates, same as create, so the next 30s reconcile turns the archive back on.

cazorla19:
Follow-up for the new e2e / helm-e2e failures on `e692eaa`:

- **e2e-tests Alert create**: connector/preset still used `time.Now().Unix()`. Operator logs were `Connector with name slack-connector-for-alert-… already exists` and `slack-connector-updated already exists`. All remaining second-granularity and hardcoded `*-updated` backend names now go through `uniqueName`.
- **helm-e2e GlobalRouter update**: create succeeded, then the backend rule name never became the patched value. The team router is account-wide; a sibling job overwrote it and there was no 30s reconcile to push the desired name back. GlobalRouter (and Preset) now get the same periodic reconcile as Connector/Scope, and the update assertion uses `accountWideBackendTimeout`.

.github/workflows/e2e-tests.yaml
```diff
@@ -44,6 +44,10 @@ jobs:
       IPACCESS_RECONCILE_INTERVAL_SECONDS: 30
       ARCHIVELOGSTARGET_RECONCILE_INTERVAL_SECONDS: 30
       ARCHIVEMETRICSTARGET_RECONCILE_INTERVAL_SECONDS: 30
+      # GlobalRouter is also account-wide: another job can overwrite the team's
+      # router while this job is asserting on an update.
```
assafad1: Router isn't account wide

internal/controller/coralogix/v1alpha1/archivemetricstarget_controller.go
```diff
@@ -95,20 +73,39 @@ func (r *ArchiveMetricsTargetReconciler) HandleCreation(ctx context.Context, log
...
+		return fmt.Errorf("error on configuring remote archivemetricstarget: %w", cxsdk.NewAPIError(httpResp, err))
+	}
+
 	updateRequest, err := archiveMetricsTarget.Spec.ExtractUpdateRequest()
 	if err != nil {
 		return fmt.Errorf("error on extracting update archivemetricstarget request: %w", err)
 	}
-	log.Info("Updating remote archivemetricstarget", "archivemetricstarget", utils.FormatJSON(updateRequest))
-	_, httpResp, err := r.ArchiveMetricsTargetsClient.
+	_, httpResp, err = r.ArchiveMetricsTargetsClient.
```
assafad1: Why are these changes needed? I thought you're fixing tests in this PR

assafad1 (approved):

PR #584, opened 2026-09-04
CX-56589: add ConfigurationGroup CRD for Fleet Manager

cazorla19:
## Summary

- Add `ConfigurationGroup` (`configurationgroups.coralogix.com`) with a nested spec: group + family + remote OpenTelemetry Collector YAML.
- Reconcile through the Preview OpenAPI client (vendored until a tagged SDK release). Delete deactivates an active family, then archives. NotFound on archive is success.
- Samples, e2e, generated CRDs, Helm chart CRDs, RBAC, and API docs included.
- `OperatorVersion` stays **2.8.0**.

Facade sync already landed the generated client on SDK master as `go/openapi/gen/fleet_manager_configuration_groups`. This PR vendors a copy; a later SDK bump can switch the import.

## Test plan

- [x] `make generate && make manifests` (generated CRDs/RBAC/docs in this PR)
- [x] `make lint && make helm-sync-check && make unit-tests` (CI on this branch)
- [x] Live mapping against EU2 Preview REST (create / replace description / get / deactivate+archive). No cluster e2e: no disposable cluster with this CRD installed.
- [x] Confirm `OperatorVersion` stays 2.8.0


cazorla19:
@codex review

cazorla19:
@codex review

internal/controller/coralogix/v1alpha1/configurationgroup_controller.go
```diff
@@ -0,0 +1,255 @@
...
+	if tags == nil {
+		tags = []string{}
+	}
+	replace.SetTags(tags)
+	priorityOrder := int32(0)
+	if group.Spec.PriorityOrder != nil {
+		priorityOrder = *group.Spec.PriorityOrder
+	}
+	replace.SetPriorityOrder(priorityOrder)
+	replace.SetFamily(*expandFamilyReplace(group.Spec.Family))
```
markshleifer-coralogix, line 179: `HandleUpdate` always sends `family` on replace (this line). The Terraform resource for the same API omits `family` when only group fields change, because a PUT with an unchanged family can mint a new family version.

The shared reconciler calls `HandleUpdate` after create (status write), on a description-only patch, and on resync. The e2e update test only checks that the group `status.id` stays the same, so family id/version churn would pass.

Can you GET after create and after a description-only replace and confirm `families[0].id` / `version` are stable? If they move, we should skip `SetFamily` when family spec did not change.

cazorla19: Checked this against the public replace handler with the payload this controller actually sends (family always present, including remotes, empty family description, empty metadata).

Sequence: create → GET → description-only PUT → GET → same PUT again (resync) → GET.

`families[0].id`, `families[0].version`, and `remoteConfigurations[0].id` were unchanged on every step. Group description did update.

The API already no-ops an unchanged family; a new family version is minted only when collector version, metadata, or remotes actually change. Terraform still omits `family` because computed IDs live in state. This CR does not store those IDs, so we should keep sending `family` on replace.

Covered in fleet-manager `TestConfigurationGroupService_ReplacePublicConfigurationGroup/keeps_family_ids_after_create,_description-only_replace,_and_resync_with_family_resent`.

cazorla19: https://github.com/coralogix/fleet-manager/pull/495

markshleifer-coralogix (approved):

PR #585, opened 2026-09-04
ci(e2e): harden ephemeral-team provisioning

cazorla19:
Follow-up to #577 (ephemeral-team isolation for the live e2e suites), hardening the rollout now that the harness lives on `main`. Part of internal ticket CX-56593.

## What changed

1. **No more bootstrap fallback.** The "Provision an ephemeral Coralogix team" step in all three workflows now always extracts the harness (`go.mod`/`go.sum`/`tests/ephemeralteam`) from `origin/main` into `$RUNNER_TEMP` and fails hard if it is missing — the PR checkout is never executed with `CORALOGIX_ORG_API_KEY` in its environment. (The fallback existed only to bootstrap #577 before the harness was on `main`.) The residual risk stays as documented: PR code still receives the minted TEAM-scoped key, the same trust class as the existing `CORALOGIX_API_KEY` secret.

2. **Group spec now runs in ephemeral teams.** `create` seeds the two fixture users the Group spec resolves by name (SCIM users client, authenticated with the minted team key, mirroring the operator's own client construction in `cmd/main.go`), and the blanket `skipIfEphemeralTeam` guard in `tests/e2e/group_test.go` is replaced by a fixture-aware gate (the spec runs wherever the member users actually resolve, and skips with the reason where they do not — including this PR's own e2e runs, since the provision step deliberately executes the harness from `origin/main`, where the seeding lands only on merge). The minted key additionally gets `team-scim:{Manage,ReadConfig}` (names verified against the SDK permission catalog). Seeding is best-effort with retries; on persistent failure it prints a loud warning and lets the Group spec fail visibly rather than aborting provisioning.

3. **Best-effort team deletion.** `delete` now disambiguates delete errors with a `TeamServiceGetTeam` — a NotFound team is already gone — and, when retries are exhausted, logs a loud warning and exits 0 instead of failing an otherwise-green job. This exact failure happened live: the org quota ledger rejected the quota release with a 400 ("DailyQuota must be greater than 0.01"). Leaked teams keep the `cx-operator-e2e-ephemeral-` name prefix and are removed by the scheduled sweeper.

## Verification

- `go build ./...`, `go vet ./...`, `gofmt`, `make lint` (0 issues), workflow YAML + actionlint clean.
- Watching the live e2e runs on this PR: expect no fallback warning in the provision step, the Group spec passing on the ephemeral team, and delete-on-success (or warn-and-keep) at the end.

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -91,20 +91,14 @@ jobs:
         run: |
           # Run the harness from the trusted base branch (origin/main), never from the PR
           # checkout, so pull-request code cannot execute with the org-wide key in its
-          # environment. Until the harness has landed on origin/main, fall back to the PR
-          # checkout with a loud warning; after the first merge the fallback is dead code.
+          # environment. The extraction fails hard if the harness is missing on origin/main.
           # Residual risk: the minted TEAM-scoped key is still handed to PR code afterwards —
           # the same trust class as today's CORALOGIX_API_KEY secret; only the ORG key is
           # confined to trusted code.
           git fetch --depth=1 origin main
-          harness_dir="$PWD"
-          if git cat-file -e origin/main:tests/ephemeralteam/main.go 2>/dev/null; then
-            harness_dir="$RUNNER_TEMP/ephemeralteam-trusted"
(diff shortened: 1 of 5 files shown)


cazorla19:
@codex review

markshleifer-coralogix (approved):

PR #588, opened 2026-09-07
chore(group): resolve members through Users OpenAPI instead of SCIM

markshleifer-coralogix:
## Summary
- Group is the only SCIM caller. Member emails are now resolved with Identity WhoAmI and Users Search, then written on the existing Team Groups OpenAPI.
- The Group CRD is unchanged. Existing manifests keep working.
- Default e2e now checks remote members. A dedicated upgrade job creates a Group on the last released (SCIM) operator, rolls to this build, and asserts the same remote id and members.

## Test plan
- [x] `make lint` and `make unit-tests`
- [x] Group e2e: create with `example@coralogix.com` / `example2@coralogix.com`, assert members, rename, delete
- [x] `group-scim-migration` job: Group created on Helm `1.0`, survives image upgrade, then update and delete
- [x] Confirm the operator API key can call WhoAmI and SearchUsers (no SCIM role required)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/upgrade-test.yaml
+++ b/.github/workflows/upgrade-test.yaml
@@ -108,12 +108,69 @@ jobs:
         run: |
           scripts/collect-logs.sh
 
+  group-scim-migration:
+    name: Group SCIM to OpenAPI migration
+    runs-on: ubuntu-latest
+    env:
+      IMG_REPO: coralogix-operator-image
+      IMG_TAG: 0.0.1
+      CORALOGIX_REGION: ${{ secrets.CORALOGIX_REGION }}
+      CORALOGIX_API_KEY: ${{ secrets.CORALOGIX_API_KEY }}
+      TEST_TEAM_ID: ${{ secrets.TEST_TEAM_ID }}
+      E2E_PROCS: "1"
+      E2E_GINKGO_LABEL_FILTER: scim-migration
+      E2E_GROUP_SCIM_MIGRATION: "1"
+      E2E_UPGRADE_IMAGE: coralogix-operator-image:v0.0.1
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Cache Makefile tool binaries
+        uses: actions/cache@v4
+        with:
+          path: bin
+          key: ${{ runner.os }}-tools-${{ github.workflow }}-${{ hashFiles('Makefile', 'go.sum') }}
+      - name: Create Kind cluster
+        uses: helm/kind-action@v1.4.0
+      - name: Install Prometheus Operator CRDs
+        run: |
+          make install-prometheus-crds
+      - name: Install Helm
+        run: |
+          curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash
+      - name: Add Coralogix Helm repository
+        run: |
+          helm repo add coralogix https://cgx.jfrog.io/artifactory/coralogix-charts-virtual
+          helm repo update
+      - name: Install the last released operator
+        run: |
+          helm install coralogix-operator coralogix/coralogix-operator \
+            --version=1.0 \
+            --namespace coralogix-operator-system \
+            --create-namespace \
+            --set secret.data.apiKey="${{ secrets.CORALOGIX_API_KEY }}" \
+            --set coralogixOperator.region="${{ secrets.CORALOGIX_REGION }}"
+      - name: Build PR image
+        uses: ./.github/actions/build-operator-image
+        with:
+          image: ${{ env.IMG_REPO }}:v${{ env.IMG_TAG }}
+      - name: Load PR image into Kind
+        run: |
+          kind load docker-image ${{ env.IMG_REPO }}:v${{ env.IMG_TAG }} --name chart-testing
+      - name: Run Group SCIM migration e2e
+        run: |
+          make e2e-tests
+      - name: Collect Operator Logs
+        if: always()
+        run: |
+          scripts/collect-logs.sh
+
   notify-failure:
     name: Notify Slack on failure
     # Scheduled (nightly) runs alert. Pull request and push runs stay quiet.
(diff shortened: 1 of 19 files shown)


api/coralogix/v1alpha1/group_types.go
```diff
@@ -55,10 +53,16 @@ type GroupSpec struct {
...
-	// Custom roles applied to the group.
+	// Custom role applied to the group.
 	// +optional
 	CustomRole *GroupCustomRole `json:"customRole,omitempty"`
 
+	// Deprecated: use customRole. Kept so Groups created against Helm chart 1.0
+	// (spec.customRoles) keep their role after CRD upgrade. The operator uses
+	// customRole when set, otherwise the first customRoles entry.
+	// +optional
+	CustomRoles []GroupCustomRole `json:"customRoles,omitempty"`
```
assafad1, line 64: Maybe I'm confused, but why are we adding a deprecated field?

markshleifer-coralogix: It's not directly related to this migration.
Helm 1.0 had `customRoles`. PR #477 switched to `customRole` and deleted the old field. In this repo, renamed fields are usually kept and marked deprecated (for example GlobalRouter fallback → fallbackTargets).
I added a Helm 1.0 upgrade test. It failed because stored Groups still have `customRoles`, and the operator no longer read that field. This PR puts `customRoles` back as deprecated and uses the first entry when `customRole` is unset.
The new logic is:
- `customRole` set → use that. Ignore `customRoles`.
- `customRole` empty, `customRoles` present → use the first list entry.
- Both empty → no role (id 0).

assafad1: We shouldn't add fields just to make tests passing IMO

.github/workflows/upgrade-test.yaml
```diff
@@ -108,12 +108,69 @@ jobs:
...
+        run: |
+          make install-prometheus-crds
+      - name: Install Helm
+        run: |
+          curl https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | bash
+      - name: Add Coralogix Helm repository
+        run: |
+          helm repo add coralogix https://cgx.jfrog.io/artifactory/coralogix-charts-virtual
+          helm repo update
+      - name: Install the last released operator
```
assafad1, line 144: is it the last one? I see `version=1.0`

.github/workflows/upgrade-test.yaml
```diff
@@ -108,12 +108,69 @@ jobs:
...
+            --version=1.0 \
+            --namespace coralogix-operator-system \
+            --create-namespace \
+            --set secret.data.apiKey="${{ secrets.CORALOGIX_API_KEY }}" \
+            --set coralogixOperator.region="${{ secrets.CORALOGIX_REGION }}"
+      - name: Build PR image
+        uses: ./.github/actions/build-operator-image
+        with:
+          image: ${{ env.IMG_REPO }}:v${{ env.IMG_TAG }}
+      - name: Load PR image into Kind
```
assafad1: don't we miss the actual upgrade step? I think you simply run e2e tests against released version here

tests/e2e/group_scim_migration_test.go
```diff
@@ -0,0 +1,356 @@
...
+				},
+			},
+			"scope": map[string]interface{}{
+				"resourceRef": map[string]interface{}{"name": scopeName},
+			},
+		},
+	}}
+}
+
+func helmUpgradeToOpenAPIOperator(ctx context.Context, image string) {
```
assafad1, line 179: can we do it in the github action, just like installation?
To be consistent with other upgrade tests.

internal/controller/coralogix/v1alpha1/group_users.go
```diff
@@ -0,0 +1,136 @@
...
+const userSearchPageSize = int64(100)
+
+// teamIDCache stores the WhoAmI team id for the process lifetime. The operator
+// reads one API key at start; a new key requires a restart.
+type teamIDCache struct {
+	mu sync.Mutex
+	id *int64
+}
+
+func (c *teamIDCache) get(ctx context.Context, identityClient *identity.IdentityServiceAPIService) (int64, error) {
```
assafad1, line 40: I really don't like that we complicate our implementation so much, to cover for a bad API.
Why does the API require a team ID? Can't they figure out the team based on the API key, like every other API?

api/coralogix/v1alpha1/group_types.go
markshleifer-coralogix: This is an actual bug the test caught. We can say that this version is too old to support, but if we want to support the upgrade from 1.0 I don't see other way.

assafad1: What is the bug?

markshleifer-coralogix: The bug is: after upgrade from Helm 1.0, the Group keeps the old YAML in etcd (customRoles), but the operator only reads customRole.

tests/e2e/group_scim_migration_test.go
markshleifer-coralogix: The existing upgrade actiob does something different: it upgrades first, then creates CRs.
What I wanted here is to test creation in 1.0 and only then upgrade.

.github/workflows/upgrade-test.yaml
markshleifer-coralogix: The name is misleading, I would change it to 1.0.

assafad1 (approved):

internal/controller/coralogix/v1alpha1/group_users.go
markshleifer-coralogix: Fixed, the new API allows no team-id calls.

assafad1 (approved):

PR #596, opened 2026-09-15
fix(reconciler): roll back remote resource when status persist fails

### LINKED ISSUE #595: Alert: Operator failing to update status.id will repeat remote Create action
# The bug

A failure to update the status of the Alert object after a successful remote apply allows for duplicated Create actions.

The generic engine `ReconcileResource` decides create-vs-update purely on !obj.HasIDInStatus(). A failure to update the object requeues the operation allowing for repeats.

# Possible solution

ReconcileResource's create branch needs to tolerate its own status-persist failing after a successful remote create, instead of blindly re-triggering HandleCreation on the next pass. 

We may need a "lock" signal into the object before API calls so we can rollback operations if the object cannot be updated, or resume the operation (not restart/repeat) in a requeue if an object update failed.

assafad1:
## Problem

Fixes #595.

`ReconcileResource` decides create-vs-update purely on `!obj.HasIDInStatus()`. After a successful remote create, it persists the returned ID into the K8s status. Today, if that status write fails it just requeues — so `HasIDInStatus()` is still `false` on the next reconcile, `HandleCreation` runs again, and a **duplicate remote resource** is created (the original is orphaned).

A delete-on-failure rollback used to guard exactly this case (added in #310) but was dropped in #472.

## Fix

Restore the rollback: if the status update after creation fails, delete the just-created remote resource so the next reconcile starts from a clean slate rather than orphaning it and creating a duplicate.

The rollback lives in the status-update failure branch, which is only reached after `HandleCreation` succeeds — so it does not interfere with the deserialization-error handling from #472 (that error is returned by `HandleCreation` and handled before the status update).

### Per-object opt-out (`CreationRollbackSkipper`)

An unconditional rollback is unsafe for some resources, so a reconciler can veto it per object:

- **imported Dashboard** — `HandleCreation` only *adopts* (looks up) the remote dashboard; deleting it would destroy a pre-existing dashboard the operator never created. Skipped when `Status.Imported` is set; a genuinely created dashboard (`Imported == false`) is still rolled back.
- **ArchiveLogsTarget / ArchiveMetricsTarget** — tenant-level singletons with a constant ID; a retry cannot duplicate, and a rollback would deactivate archiving (archive gap). Always skipped.

Default behaviour (every other reconciler) is unchanged: roll back.

## Tests

- `TestReconcileResourceCreationPersistsID` — happy path: ID persisted, finalizer added, no deletion.
- `TestReconcileResourceStatusUpdateFailureRollsBackRemoteResource` — #595: status update fails → remote resource rolled back, no stale ID.
- `TestReconcileResourceStatusUpdateFailureSkipsRollbackWhenOptedOut` — opted-out resource is **not** deleted on status-write failure.


assafad1:
@codex review

